PR #14, opened 2019-12-27
Adds conditionals for bindgen and rename use

### LINKED ISSUE #13: Make bindgen use optional
Fuchsia's build system currently does not use bindgen; let's make it possible to not have to use bindgen here too.

@kpozin

filmil:
The low-level sys crate will be compiled differently depending on
whether renaming is used in the ICU library and whether bindgen
is invoked.

The earlier version of the library had a clunkier mechanism that handles
these separate configurations.  This version is slightly less clunky,
and unifies the bindgen and not-bindgen approaches.

For the time being, the non-bindgen is only useful for ICU version 64.x.

Fixes #13.


filmil:
@kpozin Care to take a look?

rust_icu_sys/bindgen/run_bindgen.sh
```diff
@@ -0,0 +1,37 @@
...
+# This is a manual update file for bindgen.
+set -euo pipefail
+set -x
+
+# Types for which to generate the bindings.  Expand this list if you need more.
+# The syntax is regex.
+readonly GENERATE_TYPES="UBool|UCalendar.*|UChar.*|UData.*|UDate|UDateFormat.*|UEnumeration.*|UErrorCode|UText.*"
+
+# Functions for which to generate the bindings.  Expand this list if you need more.
+readonly GENERATE_FUNCTIONS="u_.*|ucal_.*|udat_.*|udata_.*|uenum_.*|uloc_.*|utext_.*"
```
kpozin: It would be great if there were some way to share these variables with `build.rs`. Barring that, we should have comments reminding developers to keep them in sync.

rust_icu_sys/Cargo.toml
```diff
@@ -23,3 +23,15 @@ paste = "0.1.5"
...
+[features]
+default = []
+
+# If set, cargo will not run bindgen, but will use the provided static libraries
+# instead.
+disable_bindgen = []
+
+# If set, ICU library version renaming is *not* performed.
+# See http://userguide.icu-project.org/design for a discussion of renaming.
+disable_renaming = []
```
kpozin: Might it be cleaner to make the feature names positive, and have `default = ["bindgen", "renaming"]`?

filmil: > Might it be cleaner to make the feature names positive, and have `default = ["bindgen", "renaming"]`?

Agreed, except it seems that features are additive. Am I missing something?

rust_icu_sys/bindgen/run_bindgen.sh
filmil: > It would be great if there were some way to share these variables with `build.rs`. Barring that, we should have comments reminding developers to keep them in sync.

Agreed here too.  Can we defer doing that until we have the current code up and running?

kpozin: Sure.

rust_icu_sys/Cargo.toml
kpozin: Users can put
```toml
rust_icu = { default-features = false }
```
in their Cargo.toml files, right?

filmil: I thought that would have been weird where we plan to use it.  But apparently it isn't.  Changed.

filmil:
@kpozin Changes made, PTAL?

PR #17, opened 2019-12-28
Improves error handling in build.rs

### LINKED ISSUE #16: Handle errors from `build.rs` properly in `rust_icu_sys`
`build.rs` today is peppered with `expect` and `unwrap`.  
Provide better diagnostics in case things go wrong during build.

filmil:
Old build.rs failed with very hermetic and user-unfriendly error
reports, making it hard to debug the root causes of build errors.

Such behavior is unnecessary, and this change improves it.

Fixes #16.


filmil:
@kpozin PTAL?

PR #18, opened 2019-12-30
Cleans up the command invocation a bit

filmil:
The new code is able to print somewhat nicer diagnostics.  Also,
grouped the commands on `icu-config` into an `impl`, to clean
the script up a bit conceptually.


filmil:
@kpozin PTAL?

Issue #20, opened 2019-12-31
Add a build feature to skip using `icu-config` altogether


filmil:
This allows us to effectively bypass the build.rs invocation, in case the build environment is prepared differently.
For example, Fuchsia has its own custom configuration that does not depend on the system ICU library and should not be using `icu-config`.

PR #21, opened 2019-12-31
Adds an option to skip icu-config run.

filmil:
Some environments do not need the `build.rs` smarts.  Allow the users
to skip using build.rs in such environments.

Fixes: #20


filmil:
@kpozin PTAL

filmil:
O HAI!

PR #22, opened 2020-01-02
Reexports features throughout.

filmil:
This looks as if it may be required for our ability to turn certain
features *off* when compiling for targets that do not require special
settings.

I'm not super-convinced that this is the case, but the only way to
test the theory is to try it out.


filmil:
@kpozin PTAL?

PR #24, opened 2020-01-03
Overrides the default feature set everywhere.

filmil:
This will perhaps allow us to import the library in Fuchsia
and override the feature set as we need to.


filmil:
@kpozin PTAL?

filmil:
@kpozin ping?

Issue #25, opened 2020-01-03
Fix the repositories link in all `Cargo.toml` files


filmil:
The repository is not correct.

PR #26, opened 2020-01-03
Corrects the repository URLs where wrong.

filmil:
Two repository paths in `Cargo.toml` files incorrectly referred
to subdirectories, instead of top-level repository URLs.

Fixes: #25


filmil:
@kpozin PTAL

PR #27, opened 2020-01-04
 Support non-generated versioned builds

### LINKED ISSUE #23: Make `feature=icu_config` robust in face of version updates
When `icu_config` feature is turned off, there is no way for the code to detect the ICU library version in use, so at the moment it always defaults to `64`.  Which, of course, will be incorrect in the general case.

Add a way for the user to pass the desired ICU renaming version.

filmil:
Adds the ability to build bindings for a user-defined
ICU library version, and to turn off build feature detection.

Since there are already plenty of configurations that one could
build the library in, I don't expect this to be the last fixup here.
But we got to start from somewhere.

Bumps the version to 0.0.4.

Fixes #23


filmil:
@kpozin WDYT? This PR is still work in progress, so no need to approve it.  

The idea here is that we prepare bindings for new versions ahead of time, so that the build environment can explicitly select the version number.  Once we have this, we can have conditional compilation that sets the version number, and we're off to the races.

kpozin:
I think this is fine as a general approach. There might be some additional complexities around supporting bindgen in the Fuchsia tree (or checking in multiple versions of pre-generated `rust_icu_sys` code, if we determine we can't do anything in-tree).

filmil:
@kpozin PTAL when able.

I discovered a few things about compilation with `cargo` (e.g. if using `build.rs` one needs to propagate feature set manually (?!)), and made both generated and static bindings for ICU 65.1, which are included in this commit.

I think we can use the env variable to make an automatic library ICU upgrade when needed.  Assuming that we can set up the compilation environment to include `RUST_ICU_MAJOR_VERSION_NUMBER` env variable, that is.

PR #29, opened 2020-01-06
Add wrapper for `uloc_acceptLanguage`

kpozin:
Note that due to a [bug in ICU4C](https://unicode-org.atlassian.net/projects/ICU/issues/ICU-20931), this only accepts exact matches.

### DIFF
--- a/rust_icu_uloc/src/lib.rs
+++ b/rust_icu_uloc/src/lib.rs
@@ -13,14 +13,25 @@
 // limitations under the License.
 
 use {
-    rust_icu_common as common, rust_icu_sys::versioned_function, rust_icu_sys::*,
-    std::convert::From, std::convert::TryFrom, std::ffi, std::os::raw,
+    rust_icu_common as common,
+    rust_icu_sys::versioned_function,
+    rust_icu_sys::*,
+    rust_icu_uenum::Enumeration,
+    std::{
+        convert::{From, TryFrom, TryInto},
+        ffi,
+        os::raw,
+    },
 };
+
+/// Maximum length of locale supported by uloc.h.
+/// See `ULOC_FULLNAME_CAPACITY`.
+const LOCALE_CAPACITY: usize = 158;
 
 /// A representation of a Unicode locale.
 ///
 /// For the time being, only basic conversion and methods are in fact implemented.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct ULoc {
     // A locale's representation in C is really just a string.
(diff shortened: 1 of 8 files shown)


kpozin:
@filmil PTAL

filmil:
Until the bug is fixed, you can work around the bug by artificially inserting a long string locale into `accepted_locales`, say if you have `nl`, you also add `nl_NL`.

rust_icu_uloc/src/lib.rs
```diff
@@ -101,6 +112,80 @@ impl ULoc {
...
+        let mut status = common::Error::OK_CODE;
+
+        let mut accept_list_cstrings: Vec<ffi::CString> = vec![];
+        // This is mutable only to satisfy the missing `const`s in the ICU4C API.
+        let mut accept_list: Vec<*const raw::c_char> = accept_list
+            .into_iter()
+            .map(|item| {
+                let uloc: ULoc = item.into();
+                accept_list_cstrings.push(uloc.as_c_str());
+                accept_list_cstrings.last().expect("non-empty list").as_ptr()
```
filmil: I wonder if we should remove the method as_c_str from `ULoc`, and introduce an as_ref instead, to avoid the issue with not recording the temporary?

That should make it impossible to step onto the issue due to lifetime requirements, and whoever needs an independent copy should clone `ULoc`.

rust_icu_uloc/src/lib.rs
```diff
@@ -228,4 +312,55 @@ mod tests {
...
+            .map(|s| ULoc::try_from(s))
+            .collect();
+        let available_locales = available_locales.expect("make available_locales");
+
+        let actual =
+            ULoc::accept_language(accept_list, available_locales).expect("call accept_language");
+        assert_eq!(
+            actual,
+            (
+                ULoc::try_from("ar_EG").ok(),
```
filmil: feel free to add your original example with two-letter locales, except you add one additional unrelated locale at end?

rust_icu_uloc/src/lib.rs
kpozin: Since we're undecided on what this API should look like, let's leave the change for a separate CL.

rust_icu_uloc/src/lib.rs
kpozin: Done

filmil:
LGTM.

Can you resolve conflicts and submit?  Let me know if unable to merge, we should make you the owner too.

kpozin:
Done

PR #34, opened 2020-01-07
Corrects the name of the docker container

filmil:
The docker container name was wrong; local tests worked because I had
the container with the same name available locally.


filmil:
@kpozin PTAL

PR #35, opened 2020-01-07
Adds some handy tools for publishing

filmil:
Adds targets `publish` and `uprev` to be used for
publication and uprevving the versions of the crates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ else
 endif
 
 # The buildenv version that will be used to build and test.
-USED_BUILDENV_VERSION := 0.0.2
+USED_BUILDENV_VERSION := 0.0.3
 
 test:
 	env LD_LIBRARY_PATH="$(shell icu-config --libdir)" cargo test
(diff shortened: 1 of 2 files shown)


filmil:
@kpozin PTAL?

Makefile
```diff
@@ -39,3 +39,47 @@ buildenv:
...
+	$(call publish,rust_icu_uenum)
+	$(call publish,rust_icu_ustring)
+	$(call publish,rust_icu_utext)
+	$(call publish,rust_icu_uloc)
+	$(call publish,rust_icu_ucal)
+	$(call publish,rust_icu_udat)
+	$(call publish,rust_icu_udata)
+
+# A helper to up-rev the cargo crate versions.
+UPREV_OLD_VERSION ?= 0.0.4
```
kpozin: Will this need to be adjusted after every version?

filmil: Yes; it does not need to get adjusted every change to the build environment; unless there is something we really want to fix ASAP.

PR #36, opened 2020-01-08
Improves the docker build-test setup

filmil:
The new setup:

- runs as the local user (instead of user root)
- mounts host source directory *and* cargo directories, so that any work
  done in the container remains usable outside of it.

This setup makes it possible to run tests inside and outside the
container, as the developer chooses to do.

The goal is the ability to run a hermetic set of tests for
different build configurations in a continuous-integration environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,21 +11,31 @@ else
 endif
 
 # The buildenv version that will be used to build and test.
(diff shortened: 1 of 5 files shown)


filmil:
@kpozin PTAL

PR #45, opened 2020-01-13
[ucal] Implement get_tz_data_version()

kpozin:


rust_icu_ucal/src/lib.rs
```diff
@@ -136,6 +136,21 @@ pub fn get_default_time_zone() -> Result<String, common::Error> {
     String::try_from(&uchar)
 }
 
+/// Implements `ucal_getTZDataVersion`
+pub fn get_tz_data_version() -> Result<String, common::Error> {
+    let mut status = common::Error::OK_CODE;
+
+    let tz_data_version = unsafe {
+        let raw_cstring = versioned_function!(ucal_getTZDataVersion)(&mut status);
+        assert!(common::Error::is_ok(status));
```
filmil: Any special reason that you prefer an `assert!` here to returning the error?

kpozin: I was following the pattern used in `set_default_time_zone` and `get_default_time_zone` above. I can abandon the pattern here.

filmil: Yes please. Likely, it's the other functions that need to be corrected to return the error instead.

filmil:
LGTM, would you mind committing once checks are done. Thx.

PR #48, opened 2020-01-14
[uloc] Implement `to_language_tag()`

kpozin:
Also:
- Implement Rust macros for generating wrappers for the numerous `uloc` methods that output strings into a buffer of a fixed size.

### DIFF
--- a/rust_icu_uloc/src/lib.rs
+++ b/rust_icu_uloc/src/lib.rs
@@ -63,6 +63,114 @@ impl TryFrom<&ffi::CStr> for ULoc {
     }
 }
 
+/// Generates a method to wrap ICU4C `uloc` methods that require a resizable output string buffer.
+///
+/// The various `uloc` methods of this type have inconsistent signature patterns, with some putting
+/// all their input arguments _before_ the `buffer` and its `capacity`, and some splitting the input
+/// arguments.
+///
+/// Therefore, the macro supports input arguments in both positions.
+///
+/// For an invocation of the form
+/// ```
+/// buffered_string_method_with_retry!(
+///     my_method,
+///     BUFFER_CAPACITY,
+///     [before_arg_a: before_type_a, before_arg_b: before_type_b,],
+///     [after_arg_a: after_type_a, after_arg_b: after_type_b,]
+/// );   
+/// ```
+/// the generated method has a signature of the form
+/// ```
+/// fn my_method(
+///     uloc_method: unsafe extern "C" fn(
+///         before_type_a,
+///         before_type_b,
+///         *mut raw::c_char,
+///         i32,
+///         after_type_a,
+///         after_type_b,
+///         *mut UErrorCode,
+///     ) -> i32,
+///     before_arg_a: before_type_a,
+///     before_arg_b: before_type_b,
+///     after_arg_a: after_type_a,
+///     after_arg_b: after_type_b
+/// ) -> Result<String, common::Error> {}
+/// ```
+macro_rules! buffered_string_method_with_retry {
+
+    ($method_name:ident, $buffer_capacity:expr,
+     [$($before_arg:ident: $before_arg_type:ty,)*],
+     [$($after_arg:ident: $after_arg_type:ty,)*]) => {
+        fn $method_name(
+            uloc_method: unsafe extern "C" fn(
+                $($before_arg_type,)*
+                *mut raw::c_char,
+                i32,
+                $($after_arg_type,)*
+                *mut UErrorCode,
+            ) -> i32,
+            $($before_arg: $before_arg_type,)*
+            $($after_arg: $after_arg_type,)*
+        ) -> Result<String, common::Error> {
+            let mut status = common::Error::OK_CODE;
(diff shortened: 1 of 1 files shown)


rust_icu_uloc/src/lib.rs
```diff
@@ -63,55 +63,178 @@ impl TryFrom<&ffi::CStr> for ULoc {
...
+
     /// Returns the current label of this locale.
     pub fn label(&self) -> &str {
         &self.repr
     }
 
     /// Returns the current locale name as a C string.
-    pub fn as_c_str(&self) -> ffi::CString {
-        ffi::CString::new(self.repr.clone()).expect("ULoc contained interior NUL bytes")
+    pub fn as_c_string(&self) -> Result<ffi::CString, common::Error> {
```
filmil: This is a backward incompatible API change, so we can not simply kill the method as that would break all users. 
 
Prefer to add a method instead, and mark this one as deprecated.

rust_icu_uloc/src/lib.rs
```diff
@@ -63,55 +63,178 @@ impl TryFrom<&ffi::CStr> for ULoc {
...
+///         after_type_b,
+///         *mut UErrorCode,
+///     ) -> i32,
+///     before_arg_a: before_type_a,
+///     before_arg_b: before_type_b,
+///     after_arg_a: after_type_a,
+///     after_arg_b: after_type_b
+/// ) -> Result<String, common::Error> {}
+/// ```
+macro_rules! buffered_string_method_with_retry {
```
filmil: I have a certain issue with macros, which is that they are difficult to maintain, and will likely remain so.  For this reason I'd like in general to use them sparingly.

Now I have to meta-compile the code in my head to understand what is going on in here.  And my editor is certainly going to fall apart trying to figure out whether there are errors in the macro source itself.

Consider the following alternatives:

1. If you must make macros, would it be possible to confine the extent of the macros only to the functions that need arg reordering, e.g. make macros that expand into closures with the correct arg binding, where the common skeleton remains the same?
2. Or make two non-macro instances of this code that do the respective correct thing for each ordering?  (Given that there is quite some variety this is probably not an option at all)

Now, I am curious to give it a test drive here, because we would benefit a lot from auto-generated binding code; so I'll approve as-is, to see where it gets us. 

But in the future please try to avoid macro use as much as possible.

rust_icu_uloc/src/lib.rs
```diff
@@ -63,55 +63,178 @@ impl TryFrom<&ffi::CStr> for ULoc {
...
     /// Implements `uloc_minimizeSubtags` from ICU4C.
     pub fn minimize_subtags(&self) -> Result<ULoc, common::Error> {
-        self.call_buffered_string_method(versioned_function!(uloc_minimizeSubtags))
+        call_buffered_string_char_star_with_repr!(self, uloc_minimizeSubtags)
             .map(|repr| ULoc { repr })
     }
 
+    // Implements `uloc_toLanguageTag` from ICU4C.
+    pub fn to_language_tag(&self, strict: bool) -> Result<String, common::Error> {
+        let repr = self.as_c_string()?;
```
filmil: Thinking aloud now.  When `as_c_string` fails,  is this a programmer error, or a user error?

I ask because it seems a bit weird that to_language_tag would need to throw an error.  This is an immutable function on `&self` which seems to indicate that we should be able to know in advance if `repr` contains a valid C string or not.  

In a way, it looks like stepping onto an error here means that our `ULoc` is already an invalid struct.  Which seems like something we should prevent.  WDYT?

rust_icu_uloc/src/lib.rs
```diff
@@ -121,19 +244,23 @@ impl ULoc {
         let mut status = common::Error::OK_CODE;
 
         let mut accept_list_cstrings: Vec<ffi::CString> = vec![];
-        // This is mutable only to satisfy the missing `const`s in the ICU4C API.
-        let mut accept_list: Vec<*const raw::c_char> = accept_list
+        let accept_list: Result<Vec<*const raw::c_char>, _> = accept_list
```
filmil: You probably want to add a comment to explain why this exists.

kpozin: Reverted.

rust_icu_uloc/src/lib.rs
kpozin: Reverted.

rust_icu_uloc/src/lib.rs
kpozin: You're right -- we don't expose any way to create an invalid `ULoc`. Reverting the `as_c_str` change.

rust_icu_uloc/src/lib.rs
kpozin: FYI, IntelliJ Rust correctly expands these macros.

Regarding **1**, so far, I haven't been able to think of a way to do this, given how many methods' signatures are variable. It might require something like [variadic generics](https://github.com/rust-lang/rfcs/issues/376). I'm open to suggestions.

As you say, **2** is impractical because we'd end up with several very long functions with almost identical code.

PR #51, opened 2020-01-17
Bump version to 0.0.5

kpozin:


kpozin:
> You may also want to bump up the version numbers in the `Makefile`
> 
> Also, you probably want to publish version 0.0.5 on crates.io.

kpozin:
Sorry, mis-clicked.

Makefile
```diff
@@ -21,7 +21,7 @@ endif
 # buildenv version by passing its value through env variables like so:
 # 
 #   make USED_BUILDENV_VERSION=whatever-you-want docker-test
-USED_BUILDENV_VERSION ?= 0.0.4
+USED_BUILDENV_VERSION ?= 0.0.5
```
filmil: You do not need to uprev buildenv version if you release new crate versions but do not touch the `build/` directory.  This value can stay at `0.0.4`

buildenv is by and large independent.  It should probably live in a separate repo to reduce the chance of confusion, but I'd rather not spend more time in OSPO reviews. :)

filmil: I think this is probably my failure to explain well the relationship between buildenv and library version -- take note, and suggestions on how to improve the explanations are most welcome.

PR #54, opened 2020-01-28
[uloc] Implement new methods and standard Rust traits for ULoc

kpozin:
- `ULoc` traits: `Hash`, `Eq`, `PartialEq`, `Ord`, `PartialOrd`
- `ULoc` methods:
  - `keywords`, `keyword_value`, `unicode_keywords`, `unicode_keyword_value`
  - `for_language_tag` (factory method)
  - Change some getters to return `Option<String>` instead of `Result<String, Error>` with possibly empty strings
- Crate functions:
  - `to_unicode_locale_key`
  - `to_unicode_locale_type`
  - `to_legacy_key`

### DIFF
--- a/rust_icu_common/src/lib.rs
+++ b/rust_icu_common/src/lib.rs
@@ -18,7 +18,7 @@
 
 use {
     rust_icu_sys as sys,
-    std::{borrow::Cow, ffi, os},
+    std::{ffi, os},
     thiserror::Error,
 };
 
@@ -39,8 +39,8 @@ pub enum Error {
 
     /// Errors originating from the wrapper code.  For example when pre-converting input into
     /// UTF8 for input that happens to be malformed.
-    #[error("wrapper error: {}", _0)]
-    Wrapper(Cow<'static, str>),
+    #[error(transparent)]
+    Wrapper(anyhow::Error),
 }
 
 impl Error {
@@ -79,12 +79,6 @@ impl Error {
         } else {
             Ok(())
         }
-    }
-
-    /// An error occurring when a string with interior NUL byte is converted to C string.
-    // TODO(fmil): rework common::Error to be more rustful.
-    pub fn string_with_interior_nul() -> Self {
-        Error::wrapper("attempted to convert a string with interior NUL byte")
     }
 
     /// Returns true if this error has the supplied `code`.
@@ -128,8 +122,26 @@ impl Error {
         }
     }
 
-    pub fn wrapper(msg: impl Into<Cow<'static, str>>) -> Self {
-        Self::Wrapper(msg.into())
+    pub fn wrapper(source: impl Into<anyhow::Error>) -> Self {
+        Self::Wrapper(source.into())
+    }
+}
+
+impl From<ffi::NulError> for Error {
+    fn from(e: ffi::NulError) -> Self {
+        Self::wrapper(e)
+    }
+}
+
+impl From<std::str::Utf8Error> for Error {
+    fn from(e: std::str::Utf8Error) -> Self {
+        Self::wrapper(e)
+    }
+}
+
+impl From<std::string::FromUtf8Error> for Error {
+    fn from(e: std::string::FromUtf8Error) -> Self {
+        Self::wrapper(e)
     }
 }
 
@@ -154,7 +166,7 @@ impl CStringVec {
         let mut rep = Vec::with_capacity(strings.len());
         // Convert all to asciiz strings and insert into the vector.
         for elem in strings {
-            let asciiz = ffi::CString::new(*elem).map_err(|_| Error::string_with_interior_nul())?;
+            let asciiz = ffi::CString::new(*elem)?;
             rep.push(asciiz);
         }
         let c_rep = rep.iter().map(|e| e.as_ptr()).collect();
--- a/rust_icu_uenum/src/lib.rs
+++ b/rust_icu_uenum/src/lib.rs
@@ -39,6 +39,17 @@ impl Enumeration {
     pub fn repr(&mut self) -> *mut sys::UEnumeration {
         self.rep
     }
+
+    /// Creates an empty `Enumeration`.
+    pub fn empty() -> Self {
+        Enumeration::try_from(&vec![][..]).unwrap()
+    }
+}
+
+impl Default for Enumeration {
+    fn default() -> Self {
+        Self::empty()
+    }
 }
 
 /// Creates an enumeration iterator from a vector of UTF-8 strings.
@@ -113,17 +124,16 @@ impl Iterator for Enumeration {
 /// Implements `ucal_openCountryTimeZones`.
 // This should be in the `ucal` crate, but not possible because of the raw enum initialization.
 // Tested in uenum.
-pub fn open_country_time_zones(country: &str) -> Result<crate::Enumeration, common::Error> {
+pub fn open_country_time_zones(country: &str) -> Result<Enumeration, common::Error> {
     let mut status = common::Error::OK_CODE;
-    let asciiz_country =
-        ffi::CString::new(country).map_err(|_| common::Error::string_with_interior_nul())?;
+    let asciiz_country = ffi::CString::new(country)?;
     // Requires that the asciiz country be a pointer to a valid C string.
     let raw_enum = unsafe {
         assert!(common::Error::is_ok(status));
         versioned_function!(ucal_openCountryTimeZones)(asciiz_country.as_ptr(), &mut status)
     };
     common::Error::ok_or_warning(status)?;
-    Ok(crate::Enumeration {
+    Ok(Enumeration {
         raw: None,
         rep: raw_enum,
     })
@@ -136,10 +146,9 @@ pub fn open_time_zone_id_enumeration(
     zone_type: sys::USystemTimeZoneType,
     region: &str,
     raw_offset: Option<i32>,
-) -> Result<crate::Enumeration, common::Error> {
+) -> Result<Enumeration, common::Error> {
     let mut status = common::Error::OK_CODE;
-    let asciiz_region =
-        ffi::CString::new(region).map_err(|_| common::Error::string_with_interior_nul())?;
+    let asciiz_region = ffi::CString::new(region)?;
     let mut repr_raw_offset: i32 = raw_offset.unwrap_or_default();
 
     // asciiz_region should be a valid asciiz pointer. raw_offset is an encoding
@@ -157,7 +166,7 @@ pub fn open_time_zone_id_enumeration(
         )
     };
     common::Error::ok_or_warning(status)?;
-    Ok(crate::Enumeration {
+    Ok(Enumeration {
         raw: None,
         rep: raw_enum,
     })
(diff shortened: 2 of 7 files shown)


rust_icu_uloc/src/lib.rs
```diff
@@ -229,6 +233,58 @@ impl ULoc {
...
+            .map(|result| result.unwrap())
+    }
+
+    /// Implements `icu::Locale::getUnicodeKeywords()` from the C++ API.
+    pub fn unicode_keywords(&self) -> impl Iterator<Item = String> {
+        self.keywords().filter_map(|s| to_unicode_locale_key(&s))
+    }
+
+    /// Implements `uloc_getKeywordValue()` from ICU4C.
+    pub fn keyword_value(&self, keyword: &str) -> Result<Option<String>, common::Error> {
```
filmil: I've been thinking... does this need to return `common::Error`?

Looks like, `keyword` could have an interior `NUL` byte technically speaking.  Could we, instead, use the potential `NulError` to re-attempt the conversion up to and including the said `NUL`?  Looks like it will make the type signature of the return value be just `Option<String>` which is nice, and I don't see that we lose anything.

Are you aware of any hidden perils of this approach?  Seems like such an approach could simplify a lot of the code below?

Also, sorry for not noticing this before --- seems like we shouldn't need to support `NUL` bytes in keywords, as the underlying ICU4C wouldn't know what to do with it anyways.  This makes it unnecessary to worry about `common::Error::string_with_interior_nul`.  `NulError` allows to [find the `NUL` byte position](https://doc.rust-lang.org/std/ffi/struct.NulError.html#method.nul_position), so perhaps we can re-attempt the conversion with this extra bit of information if it fails with `NulError` the first time around.

Thoughts?

rust_icu_uloc/src/lib.rs
```diff
@@ -267,6 +332,91 @@ impl ULoc {
...
+        let value = self.call_buffered_string_method(uloc_method).unwrap();
+        if value.is_empty() {
+            None
+        } else {
+            Some(value)
+        }
+    }
+}
+
+/// This implementation is based on ULocale.compareTo from ICU4J.
```
filmil: Why are we rolling our own?  Are there no ICU4C equivalents?

rust_icu_uloc/src/lib.rs
```diff
@@ -346,43 +496,95 @@ pub fn accept_language(
...
+        ptr.as_ref().map(|ptr| ffi::CStr::from_ptr(ptr).to_owned())
+    };
+    legacy_keyword.map(|cstring| cstring_to_string(&cstring).unwrap())
+}
+
+fn str_to_cstring(input: &str) -> Result<ffi::CString, common::Error> {
+    ffi::CString::new(input).map_err(|_| common::Error::string_with_interior_nul())
+}
+
+fn cstring_to_string(input: &ffi::CString) -> Result<String, common::Error> {
```
filmil: If we know in advance that input is a valid UTF-8 string (which we could since this is a private function), we could use the same trick as above to simplify the function signature to be just `Result<String>`.

If we don't, then looks like this should return a Result<String, Utf8Error> -- Looks to me like we're doing `unwrap` on the result anyways, so no need to convert it before discarding.

rust_icu_uloc/src/lib.rs
```diff
@@ -239,13 +295,22 @@ impl ULoc {
...
+    /// Implements `uloc_forLanguageTag` from ICU4C.
+    pub fn for_language_tag(tag: &str) -> Result<ULoc, common::Error> {
+        buffered_string_method_with_retry!(
+            buffered_string_for_language_tag,
+            LOCALE_CAPACITY,
+            [tag: *const raw::c_char,],
+            [parsed_length: *mut i32,]
+        );
+
+        let tag = ffi::CString::new(tag).map_err(|_| common::Error::string_with_interior_nul())?;
```
filmil: For example, this is one place where it's OK to leave the error conversion in place, since the constructor is allowed to produce an error.

But... it reminds me... should we have an automated `.into()` for `NulError` ? It seems possible, and the reason I didn't do this initially was mostly PEBKAC, not a real need.

rust_icu_uloc/src/lib.rs
```diff
@@ -267,6 +332,91 @@ impl ULoc {
         let asciiz = self.as_c_str();
         buffered_string_char_star(uloc_method, asciiz.as_ptr())
     }
+
+    /// Call a `uloc` method that takes this locale's ID, panics on any errors, and returns
+    /// `Some(result)` if the result is non-empty, or `None` otherwise.
+    fn call_buffered_string_method_to_option(
```
filmil: Isn't this just a [`Result::ok()`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok)?

rust_icu_uloc/src/lib.rs
```diff
@@ -229,6 +233,58 @@ impl ULoc {
...
+            ffi::CString::new(keyword).map_err(|_| common::Error::string_with_interior_nul())?;
+        buffered_string_keyword_value(
+            versioned_function!(uloc_getKeywordValue),
+            locale_id.as_ptr(),
+            keyword_name.as_ptr(),
+        )
+        .map(|value| if value.is_empty() { None } else { Some(value) })
+    }
+
+    /// Implements `icu::Locale::getUnicodeKeywordValue()` from ICU4C.
```
filmil: This could probably also benefit from the above simplification.

rust_icu_uloc/src/lib.rs
```diff
@@ -346,43 +496,95 @@ pub fn accept_language(
...
+        .map_err(|e| common::Error::Wrapper(e.to_string().into()))
+}
+
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use {super::*, anyhow::Error};
+
+    #[test]
+    fn test_language() -> Result<(), Error> {
```
filmil: Do these tests need to return a `Result` now that we've simplified the type signatures of the getters?

rust_icu_uloc/src/lib.rs
```diff
@@ -346,43 +496,95 @@ pub fn accept_language(
...
     }
 
     #[test]
-    fn test_country() {
-        let loc = ULoc::try_from("es-CO").expect("get es_CO locale");
-        let country = loc.country().expect("should get country");
-        assert_eq!(&country, "CO");
+    fn test_country() -> Result<(), Error> {
+        let loc = ULoc::try_from("es-CO")?;
+        assert_eq!(loc.country(), Some("CO".to_string()));
```
filmil: nit: should we have this implement an Eq for `Option<&str>` too?

rust_icu_uloc/src/lib.rs
```diff
@@ -421,6 +623,102 @@ mod tests {
...
+    fn test_keywords_empty() -> Result<(), Error> {
+        let loc = ULoc::for_language_tag("az-Cyrl-AZ")?;
+        let keywords: Vec<String> = loc.keywords().collect();
+        assert!(keywords.is_empty());
+        Ok(())
+    }
+
+    #[test]
+    fn test_unicode_keywords() -> Result<(), Error> {
+        let loc = ULoc::for_language_tag("az-Cyrl-AZ-u-ca-hebrew-fw-sunday-nu-deva-tz-usnyc")?;
```
filmil: Cool stuff! Thanks!

rust_icu_uloc/src/lib.rs
kpozin: I'd be okay with reattempting the `CString` conversions when there's a `NulError`.

However, in this case `uloc_getKeywordValue` can also return an error if the keyword name is ill-formed, which is a case we'd want to handle.

rust_icu_uloc/src/lib.rs
kpozin: There's nothing in the C API, and the C++ `icu::Locale` defines several operator overrides, but not `<` or `>`.

rust_icu_uloc/src/lib.rs
filmil: Aren't we swallowing that error by calling `.map(...)` in line 267?  

What am I missing?  The only place where I see an error can be returned here is `.map_err(...)` in line 261.

filmil: Oh, I guess the `.map` will leave the possible `Error` in. OK.

rust_icu_uloc/src/lib.rs
kpozin: This is checking the length of the actual `String`. I changed a comment and added a type annotation to clarify.

rust_icu_uloc/src/lib.rs
kpozin: You're right, all possible inputs to this function are coming from ICU data, so we can safely assume they're valid UTF-8.

rust_icu_uloc/src/lib.rs
kpozin: All the `ULoc` factory methods still return `Result`s, and I find using `?` cleaner than having to write explicit `.expect("this should never fail")` (or just `.unwrap()`) calls in unit tests.

rust_icu_uloc/src/lib.rs
kpozin: Do you mean another test case for the no-country case? If so, Done.

rust_icu_uloc/src/lib.rs
kpozin: Same requirement to handle invalid keywords.

rust_icu_uloc/src/lib.rs
kpozin: Added `From<NulError>` and `From<Utf8Error>` implementations for `common::Error`.

rust_icu_common/src/lib.rs
```diff
@@ -133,6 +140,18 @@ impl Error {
     }
 }
 
+impl From<ffi::NulError> for Error {
```
filmil: Thank you for adding this.

rust_icu_common/src/lib.rs
```diff
@@ -133,6 +140,18 @@ impl Error {
...
+    fn from(_: ffi::NulError) -> Self {
+        Error::string_with_interior_nul()
```
filmil: Could we use `anyhow!` and friends here to not lose the error context, by any chance?

filmil: Not necessary to add it right now, but have a though whether it would be useful.

rust_icu_uenum/src/lib.rs
```diff
@@ -168,19 +177,41 @@ pub fn open_time_zone_id_enumeration(
...
+// initialization. Tested in `uloc`.
+pub fn uloc_open_keywords(locale: &str) -> Result<Enumeration, common::Error> {
+    let mut status = common::Error::OK_CODE;
+    let asciiz_locale = ffi::CString::new(locale)?;
+    let raw_enum = unsafe {
+        assert!(common::Error::is_ok(status));
+        versioned_function!(uloc_openKeywords)(asciiz_locale.as_ptr(), &mut status)
+    };
+    common::Error::ok_or_warning(status)?;
+    // "No error but null" means that there are no keywords
```
filmil: Sigh, that API is just unnecessarily non-orthogonal.

rust_icu_common/src/lib.rs
kpozin: Done. Changed `common::Error::Wrapper` to wrap an `anyhow::Error`.

PR #55, opened 2020-02-03
Increase version to 0.1.0

kpozin:


kpozin:
This version has been published to crates.io.

PR #57, opened 2020-03-24
Adds support for basic message formatting.

filmil:
Slated for v0.1.1 since this is a backward-compatible API.

Fixes: 56

### DIFF
--- a/rust_icu_umsg/src/lib.rs
+++ b/rust_icu_umsg/src/lib.rs
@@ -1,0 +1,472 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// The message format C api is variadic.
+#![feature(c_variadic)]
+
+//! # Locale-aware message formatting.
+//!
+//! Implementation of the text formatting code from the ICU4C
+//! [`umsg.h`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/umsg_8h.html) header.
+//! Skip to the section ["Example use"](#example-use) below if you want to see it in action.
+//!
+//! The library inherits all pattern and formatting specifics from the corresponding [ICU C++
+//! API](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1MessageFormat.html).
+//!
+//! This is the support for [MessageFormat](http://userguide.icu-project.org/formatparse/messages)
+//! message formatting.  The `MessageFormat` uses ICU data to format text properly based on the
+//! locale selected at formatter initialization.  This includes formatting dates, times,
+//! currencies, and other text.
+//!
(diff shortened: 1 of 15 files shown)


filmil:
@kpozin This is work in progress.  I'll need to figure out how to get a nice API from rust, since the varargs formatter in C does not translate well.
Ideas for a type-safe formatter?  Something-something macros?

rust_icu_umsg/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+/// The formatting arguments to be used in the call to `UMessageFormat::format`.
+///
+/// According to the [ICU documentation for `umsg_format`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/umsg_8h.html#a90a4b5fe778754e5da52f7c2e5fd6048):
+///
+/// > for all numeric arguments double is assumed unless the type is explicitly
+/// > integer (long).  All choice format arguments must be of type double.
+///
+/// Additionally, we determined by inspection that the string format is necessarily
+/// `rust_icu_ustring::UChar`.
+pub enum FmtArg {
```
kpozin: Please implement `From<f64>`, `From<i32>`, etc., for `FmtArg`. For the string variants, please implement conversions from `&str` and `String`.

rust_icu_umsg/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        common::Error::ok_or_warning(status)?;
+        parse_ok(parse_status)?;
+        Ok(UMessageFormat { rep })
+    }
+
+    /// Formats a new message based on this formatter.
+    ///
+    /// Implements `umsg_format`.
+    /// Implements `umsg_vformat`.
+    pub fn format(&self, args: &[FmtArg]) -> Result<String, common::Error> {
```
kpozin: `args` should be `&[impl Into<FmtArg>]` or something similar.

rust_icu_umsg/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+) -> Result<String, common::Error> {
+    const CAP: usize = 1024;
+    let mut status = common::Error::OK_CODE;
+    let mut result = ustring::UChar::new_with_capacity(CAP);
+
+    let total_size = {
+        args.with_copy(|va_list| {
+            // Overlay a __va_list_tag on top of our va_list.
+            //
+            // This is terribly unsafe, and hinges on the assumption that the memory layout of
```
kpozin: Nit: column width

rust_icu_umsg/src/lib.rs
filmil: I rethought the formatting, and looks like `FmtArg` is not needed.

rust_icu_umsg/src/lib.rs
filmil: See above: not needed.

rust_icu_umsg/src/lib.rs
filmil: Hm, 100 cols isn't so bad.  I'd keep?

rust_icu_umsg/src/lib.rs
```diff
@@ -0,0 +1,472 @@
...
+///   )?;
+///
+///   let fmt = umsg::UMessageFormat::try_from(&msg, &loc)?;
+///   let hello = ustring::UChar::try_from("Hello! Добар дан!")?;
+///   let _ = message_format!(&fmt);
+///   let result = umsg::message_format!(
+///     fmt,
+///     { 43.4 => Double },
+///     { 31337 => Integer},
+///     { hello => String},
```
kpozin: As a future improvement, we might want to support `&str` literals here.

filmil: I am still not sure about this API... I guess we will see.

PR #67, opened 2020-03-27
Use pkg config

filmil:
Test build.


filmil:
Don't review this one, I just tried it out.

PR #69, opened 2020-03-27
Releases v0.1.2

filmil:
A patch release that should test whether it is now
possible to compile in docs.rs.


filmil:
Ping?  This fixes doc generation on `docs.rs`.

Issue #59, opened 2020-03-25
ICU 67.1 brings in new C fallback method in the API


filmil:
Between 66.0.1 and 67.1 there was a change in how locale fallback works in the ICU4C API, which fixed a long-standing bug
with the library: https://unicode-org.atlassian.net/browse/ICU-20931

Now that the bug is fixed, our language matching tests no longer work, we should fix that.

```
---- tests::test_accept_language_exact_match stdout ----
thread 'tests::test_accept_language_exact_match' panicked at 'assertion failed: `(left == right)`
  left: `(Some(ULoc { repr: "es_MX" }), ULOC_ACCEPT_FALLBACK)`,
 right: `(Some(ULoc { repr: "ar_EG" }), ULOC_ACCEPT_VALID)`', rust_icu_uloc/src/lib.rs:774:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

filmil:
@kpozin something for you?

filmil:
Note that 67.1 is not yet released. Perhaps we should just modify our instructions to check out ICU not-at-master?

filmil:
Looks like 67.1 is released.

filmil:
FYI this is now fixed with #76

PR #75, opened 2020-05-08
Fixes broken test behaviors on old and new ICUs

filmil:
With the following changes, we should be well positioned to tackle
ICU 67.1

- ICU4C correctly implements language fallbacks starting version 67.1,
  added two flavors of the same test, one for ICU versions 67 and
  onwards, and another for older versions.  Feature:
  `icu_version_67_plus`.  This happens in issue #59.
- ICU4C version 63 has wrong result for a certain test, possibly due to
  the CLDR version on my machine, so test failed.  Turned that test off
  for versions less than 64.  This happens on my machine only, when the
  system ICU is used.

See also issue #76 for where this is going to start being used.

### DIFF
--- a/rust_icu_uloc/build.rs
+++ b/rust_icu_uloc/build.rs
@@ -1,0 +1,109 @@
+#![feature(try_trait)]
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


rust_icu_uloc/build.rs
```diff
@@ -0,0 +1,109 @@
+#![feature(try_trait)]
```
kpozin, line 1: Where is this used?

PR #78, opened 2020-05-08
Starts testing with ICU 67.1

### LINKED ISSUE #76: Start testing `rust_icu` with ICU 67.1
ICU 67.1 has been [released](https://unicode-org.github.io/icu-docs/#/icu4j/com/ibm/icu/util/ULocale.html).   We should start testing with it.

A specific quirk with ICU 67.1 is that it fixes the bugs in locale matching that ICU4C had prior to this release.  See issue #59 for details.  We used not to discriminate test results based on versions, but if
we want continuity of support, we must test all supported ICU versions.  This means introducing features that allow us to run different versions of the same test depending on the version of ICU in use.

Now that we have https://github.com/google/rust_icu/pull/75, we can start setting up the build.

filmil:
This is the last step in getting rust_icu tested with ICU 67.1  There
are a few pull requests in flight that preempt this one, such as #75,
but once those are commited, this can go in. (And should pass.)

Fixes #76.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   - DOCKER_TEST_ENV=rust_icu_testenv-64
   - DOCKER_TEST_ENV=rust_icu_testenv-65
   - DOCKER_TEST_ENV=rust_icu_testenv-66
+  - DOCKER_TEST_ENV=rust_icu_testenv-67
 
 script:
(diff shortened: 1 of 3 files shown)


filmil:
This also requires build, tag, and push of the appropriate test docker images.

PR #82, opened 2020-05-08
Implement time zone methods in UCalendar

### LINKED ISSUE #73: Add time-zone related methods to `UCalendar`
There's been a request to expose functionality to get time zone offsets for a given time zone. The direct method, [`icu::TimeZone::getOffset()`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1TimeZone.html#afcbc1c48bf0b453b0123c4cb75d20e96), does not seem to have an equivalent in C, but a workaround can be cobbled together out of a few methods in `ucal.h`.

kpozin:
Fixes #73.

### DIFF
--- a/rust_icu_ucal/src/lib.rs
+++ b/rust_icu_ucal/src/lib.rs
@@ -90,6 +90,116 @@ impl UCalendar {
     pub fn as_c_calendar(&self) -> *const sys::UCalendar {
         self.rep
     }
+
+    /// Sets the calendar's current date/time in milliseconds since the epoch.
+    ///
+    /// Implements `ucal_setMillis`.
+    pub fn set_millis(&mut self, date_time: sys::UDate) -> Result<(), common::Error> {
+        let mut status = common::Error::OK_CODE;
+        unsafe {
+            versioned_function!(ucal_setMillis)(self.rep, date_time, &mut status);
+        };
+        common::Error::ok_or_warning(status)
+    }
+
+    /// Gets the calendar's current date/time in milliseconds since the epoch.
+    ///
+    /// Implements `ucal_getMillis`.
+    pub fn get_millis(&self) -> Result<sys::UDate, common::Error> {
+        let mut status = common::Error::OK_CODE;
+        let millis = unsafe { versioned_function!(ucal_getMillis)(self.rep, &mut status) };
+        common::Error::ok_or_warning(status)?;
+        Ok(millis)
+    }
+
+    /// Sets the calendar's current date in the calendar's local time zone.
+    ///
+    /// Note that `month` is 0-based.
+    ///
+    /// Implements `ucal_setDate`.
+    pub fn set_date(&mut self, year: i32, month: i32, date: i32) -> Result<(), common::Error> {
+        let mut status = common::Error::OK_CODE;
+        unsafe {
+            versioned_function!(ucal_setDate)(self.rep, year, month, date, &mut status);
+        }
+        common::Error::ok_or_warning(status)?;
+        Ok(())
+    }
+
+    /// Sets the calendar's current date and time in the calendar's local time zone.
+    ///
+    /// Note that `month` is 0-based.
+    ///
+    /// Implements `ucal_setDateTime`.
+    pub fn set_date_time(
+        &mut self,
+        year: i32,
+        month: i32,
+        date: i32,
+        hour: i32,
+        minute: i32,
+        second: i32,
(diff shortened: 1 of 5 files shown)


filmil:
Looking.

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
         let tz_version = super::get_tz_data_version().expect("get_tz_data_version works");
         assert!(re.is_match(&tz_version));
     }
+
+    #[test]
+    fn test_get_set_millis() -> Result<(), common::Error> {
+        let now = get_now();
+        let mut cal = UCalendar::new("America/New_York", "en-US", UCalendarType::UCAL_GREGORIAN)?;
+        assert!((now - cal.get_millis()?).abs() < 1000f64);
```
filmil, line 339: Care to explain this magic check?

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
...
+
+        let date = now + 17.0;
```
filmil, line 341: Declare a clearly named constant instead of magic 17?

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
...
+        cal.set_millis(date)?;
+        assert_eq!(cal.get_millis()?, date);
+        Ok(())
+    }
+
+    #[test]
+    fn test_set_date() -> Result<(), common::Error> {
+        // Thu May 07 2020 21:00:00 GMT-0400 (US Eastern Daylight Time)
+        let time_a: UDate = 1588899600000.0;
```
filmil, line 350: Why not use parse time?  I think calendar supports it?

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
...
+        assert_eq!(cal.get(UCalendarDateFields::UCAL_MILLISECOND)?, 898);
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_offsets_and_daylight_time() -> Result<(), common::Error> {
+        let mut cal = UCalendar::new("America/New_York", "en-US", UCalendarType::UCAL_GREGORIAN)?;
+
+        let expected_zone_offset: i32 = -5 * 60 * 60 * 1000;
```
filmil, line 398: 5 hours vs 1 hour?  Units? Document?  Here and below.

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
...
+        let expected_dst_offset: i32 = 1 * 60 * 60 * 1000;
+
+        cal.set_date_time(2020, 4 /* May */, 7, 21, 0, 0)?;
```
filmil, line 401: Nit: would it be worthwhile to define separate enum types for each of these params?  They are kind of easy to mix up at the moment. 0 for January is especially weird.

rust_icu_ucal/src/lib.rs
```diff
@@ -217,4 +331,86 @@ mod tests {
...
+    #[test]
+    fn test_set_date_time() -> Result<(), common::Error> {
+        // Thu May 07 2020 21:26:55.898 GMT-0400 (US Eastern Daylight Time)
+        let time_a: UDate = 1588901215898.0;
+        // Mon May 04 2020 21:00:00.898 GMT-0400 (US Eastern Daylight Time)
+        let time_b: UDate = 1588640400898.0;
+
+        let mut cal = UCalendar::new("America/New_York", "en-US", UCalendarType::UCAL_GREGORIAN)?;
+        cal.set_millis(time_a)?;
+        cal.set_date_time(2020, 4 /* May */, 4, 21, 0, 0)?;
```
filmil, line 371: Consider also adding a test that is always about a date which is 1 year ahead of whatever is "today".  This way, if things somehow change such that we'd be broken in the future, we'd know 1 year in advance.

filmil: > Do you mean something like a test that confirms that America/New_York still follows DST one year in the future?

This, in fact.  But, at your option.

kpozin: We don't have any code (to my knowledge) that relies on time zone rules remaining static, so let's avoid creating tests that might break unhelpfully :)

PR #88, opened 2020-05-12
Starts using github workflows

filmil:
Travis has for some reason stopped reporting build
statuses to github.  Which makes it way less useful
than before.

I'll leave the travis integration around for a bit until
we're sure about github integration.

Fixes: 84


filmil:
@kpozin : Github actions may be better supported and less hassle than Travis.  We can vote with our feet.

This change does not require holding a release back.

Issue #96, opened 2020-05-13
Amend `make publish` to support removing dev-dependencies


kpozin:
crates.io validates all dependencies, including dev-dependencies. If there's an apparent cycle caused by imports in tests, e.g. [ucal] -> [udat] -> [ucal dev], it becomes impossible to publish either crate to crates.io. (See https://github.com/rust-lang/cargo/issues/4242.)

As a workaround, we could have `make publish` something along the lines of:
1. Verify that the git workspace is clean.
1. Back up all Cargo.toml files.
1. Remove or comment out `[dev-dependencies]` from all Cargo.toml files.
1. Run `cargo publish --allow-dirty`, allowing the modified workspace state.
1. Restore the backed up Cargo.toml files.

filmil:
Alternative: is it possible to somehow arrange the crates so that the circular dependency is removed altogether?

kpozin:
We could theoretically move the tests that cause the cycle out into a new crate, but that seems like a risky practice.

filmil:
To short-circuit the search, which crates are forming a cycle?  Perhaps indeed we can move things around so that we don't need to patch things up to publish them.

kpozin:
It's `ucal` and `udat`, because

* `udat` has a method that takes a `&ucal::UCalendar`

https://github.com/google/rust_icu/blob/29ff766701c5fced71b46955f57cc46acba69c8e/rust_icu_udat/src/lib.rs#L147
   
* In `ucal` tests, we're parsing dates (`udat::UDateFormat`)

https://github.com/google/rust_icu/blob/29ff766701c5fced71b46955f57cc46acba69c8e/rust_icu_ucal/src/lib.rs#L422

filmil:
In this particular case, it looks like we could move the test function [`iso_8601_format`](https://github.com/google/rust_icu/blob/29ff766701c5fced71b46955f57cc46acba69c8e/rust_icu_ucal/src/lib.rs#L417) from `ucal` to `udat`, but expose it with `#[cfg(test)]`.  Perhaps we could move other such functions over in a similar vein?

filmil:
I removed my earlier message, since I misunderstood the options we had.

So now I see and understand how this came about.  Given that `rust_icu_udat` is a higher-level crate than `rust_icu_ucal`, it is wrong for tests in `rust_icu_ucal` to require a loop back.  So I kind of side with the crates.io objecting in the end; even though strictly speaking dev deps could be treated separately.

[...]

PR #101, opened 2020-05-14
Removes circular dev dependency ucal(dev)->udat->ucal

filmil:
This change moves a few `ucal` tests up the test hierarchy
to remove a dev dependency of ucal onto udat.  This circular
dependency makes crates.io object to the dependency structure,
even though strictly speaking a dev dependency shouldn't count.

See issue #96 for a discussion of options.  I think this is the
least invasive change that we can make to keep the tests in their
current form.

Fixes: #96

### DIFF
--- a/rust_icu_ucol/Cargo.toml
+++ b/rust_icu_ucol/Cargo.toml
@@ -58,10 +58,6 @@ icu_version_in_env = [
 icu_version_64_plus = []
 icu_version_67_plus = []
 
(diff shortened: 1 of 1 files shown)


filmil:
@kpozin my proposal to remove the circular dev deps.

kpozin:
We could also go back to just hard-coding the timestamps as integers instead of parsing them.

PR #103, opened 2020-05-14
Remove dev-dependency cycle in ucal

### LINKED ISSUE #96: Amend `make publish` to support removing dev-dependencies

kpozin:
In `rust_icu_ucal`, replace `UDateFormat` parsing with hard-coded timestamps.

Closes #96 with a workaround.

### DIFF
--- a/rust_icu_ucal/src/lib.rs
+++ b/rust_icu_ucal/src/lib.rs
@@ -298,9 +298,6 @@ mod tests {
     use {
         super::{UCalendar, *},
         regex::Regex,
-        rust_icu_udat::UDateFormat,
-        rust_icu_uloc::ULoc,
-        rust_icu_ustring::UChar,
         std::collections::HashSet,
     };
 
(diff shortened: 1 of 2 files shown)


rust_icu_ucal/src/lib.rs
```diff
@@ -413,21 +410,12 @@ mod tests {
...
-        UDateFormat::new_with_pattern(&locale, &tz, &pattern)
-    }
-
     #[test]
     fn test_set_date() -> Result<(), common::Error> {
-        let date_format = iso_8601_format()?;
-
-        let time_a = date_format.parse("2020-05-07T21:00:00.000-04:00")?;
-        let time_b = date_format.parse("2020-05-04T21:00:00.000-04:00")?;
+        // 2020-05-07T21:00:00.000-04:00
```
filmil, line 415: Consider also adding a note why this is not parsed.

kpozin: Done

PR #112, opened 2020-06-06
Adds feature testing to ICU 63 and 64

filmil:
Feature testing is needed to ensure that older generated code supports
newly added `rust_icu_*` crates.

Fixes: 111


filmil:
Note @kpozin, most of the bulk of this change is version uprev.

PR #116, opened 2020-06-10
Use std::os::raw::c_char instead of hardcoded i8

betamos:
This was detected in a build failure on Fuchsia on arm64.


filmil:
Thank you!

PR #122, opened 2020-06-10
Start using buildenv 1.1.0

### LINKED ISSUE #120: Build workflow with features is not set up correctly
Some feature combinations are never actually exercised, like "64 plus" and "67 plus". We are effectively just not running the respective tests.
:(

filmil:
Buildenv 1.1.0 is hopefully upgraded to do *correct* feature testing.
The answer here seems to be "yes, it's correct", based on spot-checks.

One more issue came up while I was probing: the builds that didn't use
the build script didn't get the appropriate `-L` flag added to the
linker command, and would therefore fail.  I added the .cargo/config
file for that.  While in theory these settings could clash with those
produced by the build.rs, I think in practice this will likely be rare. (You'd
need to have a conflicting version of ICU in /usr/local/lib *and* elsewhere.)

As result, I also restructured the test workflows:
- Broke the tests apart into subsets - now all ICU versions 64 and up
  are tested with the flag icu_version_64_plus.
- Removed incompatible flag combinations: `use_bindgen` is incompatible
  with `icu_version_in_env`.

Closes #120.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,8 @@
+# These flags are the only way to tell cargo to add directories to the library
+# search path.  When `build.rs` script is used, this is not needed since the
+# script will set the library paths correctly.  But when it isn't used,
+# something still needs to produce these flags.  With cargo, this is the only
+# way.
+[build]
+rustflags=["-L", "/usr/local/lib"]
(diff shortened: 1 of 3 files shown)


filmil:
@kpozin PTAL, I think the automated builds are finally in the correct configuration.

Issue #127, opened 2020-06-16
Implement `Intl.PluralRules`


filmil:
Task list:

- Define trait
- Expose `upluralrules.h`
- implement trait in terms of `upluralrules.h`

PR #128, opened 2020-06-16
Adds a trait definition for Intl.PluralRules.

filmil:
See issue #127.

### DIFF
--- a/ecma402_traits/src/pluralrules.rs
+++ b/ecma402_traits/src/pluralrules.rs
@@ -1,0 +1,97 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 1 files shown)


filmil:
@zbraniecki

ecma402_traits/src/pluralrules.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct Options {
+    /// Selects a [options::Type] for the formatted list.  If unset, defaults to
+    /// [options::Type::Cardinal].
+    pub in_type: options::Type,
+    pub minimum_integer_digits: u8,
+    pub minimum_fraction_digits: u8,
+    pub maximum_fraction_digits: u8,
+    pub minimum_significant_digits: u8,
+    pub maximum_significant_digits: u8,
```
zbraniecki, line 51: Would it make sense to consider using `std::ops::RangeBounds`? 

I feel like it makes it a bit more intuitive that we're talking ranges, rather than independent min/max values. But I'm not sure how traits will work here, so if that's a no-go, then the current layout is ok.

filmil: The idea of using a dedicated, already-existing type is appealing.  

I noticed that it's not obvious which is the concrete type implementing `RangeBounds` (it's a trait).  Also, it has a whole lot of complication with the use of inclusive and exclusive bounds, for which we don't have the need here.

Long story short, I'm not quite seeing how to get the benefit of a dedicated type without complicating the trait definition too much.  Given that, I'd opt to keep things as they are for the time being.  I'm open to reworking the options if something better comes along.

And of course, I may simply be wrong in the above assessment, in which case I welcome corrections.

zbraniecki: yeah, we can revisit it later :)

PR #130, opened 2020-06-17
ICU upluralrules

### LINKED ISSUE #127: Implement `Intl.PluralRules`

filmil:
Adds the support for locale-sensitive plural rules

Plural rules determines what gets printed when one needs to format "1 file"
or "10 files" for human consumption.  Each language may have very own rules
about what number is considered singular, what is considered plural.

Languages have different rules about the numbers subject to the rule, and
even about the number of plurals that are used.

Closes #127

### DIFF
--- a/rust_icu_uenum/src/lib.rs
+++ b/rust_icu_uenum/src/lib.rs
@@ -122,6 +122,26 @@ impl Iterator for Enumeration {
     }
 }
 
+impl Enumeration {
+    /// Constructs an [Enumeration] from a raw pointer.
+    ///
+    /// **DO NOT USE THIS FUNCTION UNLESS THERE IS NO OTHER CHOICE!**
+    ///
+    /// We tried to keep this function hidden to avoid the
+    /// need to have a such a powerful unsafe function in the
+    /// public API.  It worked up to a point for free functions.
+    ///
+    /// It no longer works on high-level methods that return
+    /// enumerations, since they we'd need to depend on them to
+    /// create an [Enumeration].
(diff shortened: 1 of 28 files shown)


filmil:
@kpozin I stacked multiple commits here, let me know if this makes the review unwieldy and I'll do differently in the future.

rust_icu_uenum/src/lib.rs
```diff
@@ -122,6 +122,26 @@ impl Iterator for Enumeration {
...
+    ///
+    /// **DO NOT USE THIS FUNCTION UNLESS THERE IS NO OTHER CHOICE!**
+    ///
+    /// We tried to keep this function hidden to avoid the
+    /// need to have a such a powerful unsafe function in the
+    /// public API.  It worked up to a point for free functions.
+    ///
+    /// It no longer works on high-level methods that return
+    /// enumerations, since they we'd need to depend on them to
+    /// create an [Enumeration].
```
kpozin, line 136: > since they we'd need to depend on them

Can this be rephrased?

filmil: ah, `s/they/then/` I suppose?

kpozin: SGTM.
Also, let's add `#[doc(hidden)]`?

filmil: Done.

PR #131, opened 2020-06-18
Implements ECMA 402 pluralrules using ICU

filmil:
Adds the `select` implementation that does not use the formatting hints
from the number formatter.

That's being left for followup work.

### DIFF
--- a/rust_icu_ecma402/src/pluralrules.rs
+++ b/rust_icu_ecma402/src/pluralrules.rs
@@ -1,0 +1,142 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


filmil:
@zbraniecki FYI

rust_icu_ecma402/src/pluralrules.rs
```diff
@@ -0,0 +1,142 @@
...
+pub struct PluralRules {
+    // The internal representation of rules.
+    rep: uplr::UPluralRules,
+}
+
+pub(crate) mod internal {
+    use ecma402_traits::pluralrules::options;
+    use rust_icu_sys as usys;
+
+    // Converts the trait style option type to an equivalente ICU type.
```
kpozin, line 33: "equivalente" is the Spanish spelling :)

filmil: Done.

Issue #141, opened 2020-06-23
Implement `unum.h`


filmil:
The function in the [`unum.h` header](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/unum_8h.html) provide number formatting.  Needed for issue #137.

PR #143, opened 2020-06-24
Implementing `umsg`

filmil:
Since the entire implementation is large, I thought to split up into smaller pieces.  Heres' the first piece, whcih in turn has a few commits to its name.

Issue #141


rust_icu_common/src/lib.rs
```diff
@@ -333,3 +336,24 @@ mod tests {
...
+/// Converts a parse error to a Result.
+pub fn parse_ok(e: sys::UParseError) -> Result<(), crate::Error> {
+    if e == NO_PARSE_ERROR {
+        return Ok(());
+    }
+    Err(Error::Wrapper(anyhow!(
+        "parse error: line: {}, offset: {}",
+        e.line,
+        e.offset
+    )))
```
kpozin, line 357: Would there be any value in printing `preContext` and `postContext`?

rust_icu_unum/src/lib.rs
```diff
@@ -20,10 +20,68 @@ use {
...
+        locale: &uloc::ULoc,
+    ) -> Result<UNumberFormat, common::Error> {
+        let mut status = common::Error::OK_CODE;
+        let mut parse = common::NO_PARSE_ERROR.clone();
+        let loc = locale.as_c_str();
+
+        // Unsafety note: all variables should be valid.
+        let rep = unsafe {
+            assert!(common::Error::is_ok(status));
+            assert!(common::parse_ok(parse).is_ok());
```
kpozin, line 88: Out of curiosity, is this really needed? It does have some (minuscule) cost, and I can't think of any circumstances in which the assertion could ever fail.

rust_icu_unum/src/lib.rs
```diff
@@ -37,10 +37,45 @@ impl Drop for UNumberFormat {
...
+    ) -> Result<UNumberFormat, common::Error> {
+        UNumberFormat::try_new_style_pattern_ustring(
+            sys::UNumberFormatStyle::UNUM_PATTERN_DECIMAL,
+            pattern,
+            locale,
+        )
+    }
+
+    /// Implements `unum_open`, with rule-based formatting,
+    pub fn try_new_decimal_rulebased_ustring(
```
kpozin, line 53: Naming nit: `...rule_based...`

rust_icu_unum/src/lib.rs
```diff
@@ -81,7 +116,158 @@ impl UNumberFormat {
...
+        for test in tests {
+            let locale = uloc::ULoc::try_from(test.locale).expect("locale exists");
+            let pattern = ustring::UChar::try_from(test.rule).expect("pattern is set");
+            let fmt = crate::UNumberFormat::try_new_decimal_rulebased_ustring(&pattern, &locale)
+                .expect("formatter");
+
+            let result = fmt
+                .try_clone()
+                .expect("clone")
+                .format(42)
```
kpozin, line 230: This value should be part of the `TestCase`, right?

rust_icu_unum/src/lib.rs
filmil: If I didn't have an assert here, I'd have a comment that says essentially the same thing.  So, why not have an assert to begin with? it's the same sort of precondition commitment, but with more teeth.

And every compiler worth its salt should be able to be told to not generate assertion code if that's needed.

rust_icu_common/src/lib.rs
filmil: IIRC the code that we have wrapped so far does not fill out those two values.

rust_icu_unum/src/lib.rs
filmil: Done.

PR #144, opened 2020-06-24
Creates a Unicode string macro with retry

filmil:
There is a macro buffered_String_method_with_retry which does this
little variable buffer dance that is sometimes needed in the ICU4C
API.  This macro is the same thing, except the end result is not a
UTF8 string, but a UChar string.

This will come in handy in Issue #141, since `unum.h` comes with 
quite a few methods that fit this general pattern.


filmil:
@kpozin not sure what I did differently this time -- everything unrolled as expected.

PR #145, opened 2020-06-25
Unum part 2

filmil:
Further down the road of implementing `unum.h`.

The general goal is to implement all but the "parse" functions, since those seem to be of niche interest as opposed to things used to display data.


rust_icu_unum/src/lib.rs
```diff
@@ -200,6 +201,73 @@ impl UNumberFormat {
         )?;
         Ok((result, iterator))
     }
+
+    /// Implements `unum_formatDecimal`. Since 0.3.1.
```
kpozin, line 206: It's unfortunate that rustc doesn't allow `#[stable(since = "0.3.1")]` outside of the standard library.

filmil:
I had to reimage my machine and lost these changes :(  Will get back to them when time allows.

PR #146, opened 2020-06-26
Unum part 3

filmil:
Part 3 in the unum number formatting support work.

Please see individual commits for details of the change.

# Adds support for UFormattable.

Using uformattable::UFormattable provides
generalized value parsing for the particular
locale.

While this isn't the full spectrum of the
underlying functions that ICU4C provides
for value parsing, it's general enough
to be usable for all the use cases where
parsing is needed, *if* needed.  However,
those uses are rare.

Issue #141.

### DIFF
--- a/rust_icu_unum/src/lib.rs
+++ b/rust_icu_unum/src/lib.rs
@@ -21,7 +21,7 @@ use {
     paste, rust_icu_common as common, rust_icu_sys as sys,
     rust_icu_sys::versioned_function,
     rust_icu_sys::*,
-    rust_icu_uloc as uloc, rust_icu_ustring as ustring,
+    rust_icu_uformattable as uformattable, rust_icu_uloc as uloc, rust_icu_ustring as ustring,
     rust_icu_ustring::buffered_uchar_method_with_retry,
     std::{convert::TryFrom, convert::TryInto, ptr},
 };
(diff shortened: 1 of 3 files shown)


rust_icu_unum/src/lib.rs
```diff
@@ -263,12 +271,83 @@ impl UNumberFormat {
...
+            number,
             currencyz.as_mut_c_ptr(),
             0 as *mut sys::UFieldPosition,
         )
     }
-   
+
+    /// Implements `unum_parseToUFormattable`. Since 0.3.1.
+    ///
+    /// > **WARNING** the `parse_position` parameter is with respect to the number index
```
kpozin, line 282: "with respect to the number index" is a bit unclear. Maybe something like the "starting index of the numeric string"?

rust_icu_unum/src/lib.rs
```diff
@@ -263,12 +271,83 @@ impl UNumberFormat {
...
+        let result = self.format_formattable_ustring(fmt)?;
+        String::try_from(&result)
+    }
+
+    /// Implements `unum_formatUFormattable`. Since 0.3.1.
+    pub fn format_formattable_ustring<'a>(
+        &self,
+        fmt: &uformattable::UFormattable<'a>,
+    ) -> Result<ustring::UChar, common::Error> {
+        const CAPACITY: usize = 200;
```
kpozin, line 333: Is this capacity recommended somewhere in ICU4C documentation? Consider mentioning the source in a comment.

PR #148, opened 2020-07-17
Unum part 4

filmil:
Multiple commits.  First one is automated.

- Adds more allowed types into the bindgen and build.rs, updates the static bindings and coverage
- Implementation proper.


rust_icu_unum/src/lib.rs
```diff
@@ -443,6 +675,42 @@ impl<'a, T> Iterator for UFieldPositionIterator<'a, T> {
...
+    fn next(&mut self) -> Option<&'static str> {
+        if self.max == 0 {
+            return None;
+        }
+        if self.next >= self.max {
+            return None;
+        }
+        let c = unsafe {
+            let cptr = versioned_function!(unum_getAvailable)(self.next as i32);
+            std::ffi::CStr::from_ptr(cptr)
```
kpozin, line 707: If `cptr` is null, this will segfault.

rust_icu_unum/src/lib.rs
```diff
@@ -443,6 +675,42 @@ impl<'a, T> Iterator for UFieldPositionIterator<'a, T> {
...
+        };
+        self.next = self.next + 1;
+        Some(c.to_str().expect("can be converted to str"))
```
kpozin, line 710: It might be worth commenting why we're confident that `'static` is the correct lifetime here.

rust_icu_common/src/lib.rs
```diff
@@ -146,6 +146,12 @@ impl From<std::string::FromUtf8Error> for Error {
     }
 }
 
+impl From<anyhow::Error> for Error {
```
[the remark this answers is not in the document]
filmil: Done. I didn't know this.

rust_icu_unum/src/lib.rs
filmil: Added an assertion and a comment. I am not sure how we can guard against this possibility.

rust_icu_unum/src/lib.rs
filmil: *Excellent* point!  

I guess, there are no guarantees.  

When writing this, I thought that this call would be returning compiled-in values (hence `'static`), but then realized this is not true when locale data is loaded from a file, and is *especially* not true if the data is reloaded during the lifetime of the iterator.

And since there are no global objects to hang the lifetime off of, this has to return `Option<String>` instead.

There are other gotchas remaining, such as any data invalidation that happens while the iterator is live.  Which I don't have a good idea how to remove.

filmil:
@kpozin PTAL, your concerns have been addressed, at least to an extent that I thought would be practical.  Thoughts?

PR #156, opened 2020-07-22
Start using ICU 67 as default for docker tests

filmil:
As we've moved Fuchsia to ICU 67, we may as well bump the version
up here.

Ideally, we'd keep this close to the newest supported ICU library
version.  We kept this on 64 to match Fuchsia; now this is no
longer needed.


filmil:
PTAL. I thought we'd bump this version indicator up.  Preparing for ICU 68.1...

PR #158, opened 2020-08-15
bump paste version 0.1 -> 1.0

ComputerDruid:


ComputerDruid:
CI might not pass right now; I had some trouble with the docker flow locally. It worked with `cargo test` though, so I think it's probably that the docker images need to be rebuilt with rust >= 1.45

filmil:
Thanks. I’ll take a look early next week.

Out of curiosity, any  special reasons for the version bump? 

Also would you mind filling a bug for the docker flow? I’m very interested in keeping it healthy.

ComputerDruid:
Pretty much just because it would be nice if http://fxrev.dev/417545 didn't have to vendor 2 versions of paste.

And I don't think the docker problem is universal; I think it's specifically that it's too old to build paste 1.0.

filmil:
Noted.  But even in that case, I'm reluctant to break the buildbots, so I'd appreciate if this pull request waited until a new release of buildenv is available, assuming the issue is indeed isolated to what you say.  (buildenv is the docker image that all build bots use)

ComputerDruid:
If by buildenv you mean https://hub.docker.com/r/filipfilmar/rust_icu_testenv/ , then yeah, absolutely. Let me know if I can help with that

filmil:
FYI see https://github.com/google/rust_icu/pull/159. 
I'll use that change to test this pull request.

filmil:
Confirmed the following:

* this pull request builds with `buildenv-1.3.0-rc.1`, new buildenv release candidate from https://github.com/google/rust_icu/pull/159
* but fails with old `buildenv-1.1.0`, with the same error as on buildbots: `error[E0658]: procedural macros cannot be expanded to expressions`

So I think that once https://github.com/google/rust_icu/pull/159 is merged, and `buildenv-1.3.0` is released, this pull request will be mergeable.  Stay tuned.

filmil:
@ComputerDruid Would you mind rebasing this pull request and trying again?

New buildenv has been deployed, and IIUC it should allow the presubmit checks to pass here.

ComputerDruid:
@filmil Yep, that seems to have done it! Thanks!

PR #160, opened 2020-08-17
Start using buildenv-1.3.0

filmil:
This starts using a newer nightly Rust toolchain (1.47)
which, in turn, can compile paste-1.0.0 from pull
request: #158


filmil:
I don't expect this to turn green until https://github.com/google/rust_icu/pull/159 is merged and `buildenv-1.3.0` is pushed to docker hub.

Issue #157, opened 2020-08-13
No Transliterate? (utrans.h)


tsukit:
Hi there. I don't see the mapping to `utrans.h` in C API, or anything that's for transliterate (http://userguide.icu-project.org/transforms/general). But I might have missed it. If there's current support for it, could you please point me to it?

filmil:
Hi tsukit@.  You haven't missed anything, rust_icu doesn't yet have complete coverage.  If you want to contribute `utrans.h`, we'd be very glad to accept the contribution.

filmil:
This is now implemented.  Thanks to @bdragon .

PR #161, opened 2020-08-26
Implement utrans.h

bdragon:
This PR adds bindings for most of [`utrans.h`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/utrans_8h.html).

I ran `make static-bindgen` to re-generate the bindgen code with the `utrans.h` APIs for each of the targeted ICU versions. I also re-generated the coverage progress report, but I only included the part of the coverage diff that was relevant to this feature. Otherwise, I tried to follow the conventions and style of existing code in `rust_icu` wherever possible.

## API

A handful of the `utrans_*` C APIs have been deprecated in favor of newer ones:

| Deprecated ICU4C API | Preferred ICU4C API |
| -------------------- | ------------------- |
| `utrans_countAvailableIDs` | `utrans_openIDs` |
| `utrans_getAvailableID` | `utrans_openIDs` |
| `utrans_getID` | `utrans_getUnicodeID` |
| `utrans_open` | `utrans_openU` |
| `utrans_transIncremental` | `utrans_transIncrementalUChars` |
| `utrans_unregister` | `utrans_unregisterID` |

I have ignored the deprecated APIs.

In an effort to avoid carrying over some of the odd naming that comes from having to maintain the deprecated APIs alongside newer ones, I have chosen to name the Rust APIs simply, even if the chosen name more closely resembles the deprecated API than the preferred API (e.g., `.get_id` rather than `.get_unicode_id`). I recognize that this is fairly subjective and I welcome pushback on the decision or other suggestions for naming.

| Preferred ICU4C API | rust_icu_utrans API |
| --------- | ------------ |
| `utrans_clone` | `impl Clone` |
| `utrans_close` | `impl Drop` |
| `utrans_getSourceSet` | Not implemented.<sup>1</sup> |
| `utrans_getUnicodeID` | `.get_id` |
| `utrans_openIDs` | `::get_ids` |
| `utrans_openInverse` | `::inverse` |
| `utrans_openU` | `::new`, `::new_ustring` |
| `utrans_register` | `::register` |
| `utrans_setFilter` | `.set_filter`, `.set_filter_ustring` |
| `utrans_toRules` | `.to_rules` |
| `utrans_trans` | Not implemented.<sup>2</sup> |
| `utrans_transIncrementalUChars` | Not implemented.<sup>3</sup> |
| `utrans_transUChars` | `.transliterate`, `.transliterate_ustring` |
| `utrans_unregisterID` | `::unregister` |

---

<sup>1</sup> `utrans_getSourceSet` returns a [`USet`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uset_8h.html), which has not been implemented in `rust_icu`. Given the area of that API, it seems out-of-scope for this feature, and so I have omitted `utrans_getSourceSet`. If this PR moves forward, I would be happy to take on the work of adding bindings for the `uset_` APIs in a future PR and circling back to add `utrans_getSourceSet`.

<sup>2</sup> `utrans_trans` relies on the [`UReplaceable` and `UReplaceableCallbacks`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/urep_8h.html) types—an opaque type and a bag of function pointers that operate on it—which, from what I can grok, seem to be the preferred way of doing things in the ICU C++ APIs. While these types _are_ exposed as part of the C interface, there does not appear to be any way to obtain them in order to pass them to `utrans_trans` without relying on C++ APIs. Given that `utrans_transUChars` provides us with the same functionality, I chose to leave `UReplaceable` and `UReplaceableCallbacks` on the shelf and implement the as-yet more familiar `UChars`-based approach.

<sup>3</sup> Having implemented `utrans_transUChars`, I didn't feel the need to implement `utrans_transIncrementalUChars`, which is distinct to `utrans_transUChars` in that it works with a `UTransPosition`, but I can to add it in this PR for completeness if requested.


filmil:
Thank you for contributing the binding for `utrans.h`, and for taking time to document your work so thoroughly. I will review the PR shortly.

filmil:
FYI, there was a feature request for exactly this at: https://github.com/google/rust_icu/issues/157

README.md
```diff
@@ -477,8 +478,9 @@ to include more features.
 
 When adding more ICU wrappers, make sure to do the following:
 
-* Check `build.rs` to add appropriate lines into `bindgen_source_modules`, then
-  `whitelist_types_regexes` and `whitelist_functions_regexes`.
+* Check `rust_icu_sys/build.rs` and `rust_icu_sys/bindgen/run_bindgen.sh` to add
```
filmil, line 481: Thank you for fixing this piece of documentation!

filmil:
Thank you for the contribution.  If you don't mind leaving some feedback on the developer experience when working on this PR (e.g. what worked, and what didn't), it would be most appreciated.

filmil:
@googlebot rescan

bdragon:
@filmil Thank you for the fast response and code review!

The README made it easy to get started and I found the code to be well-organized. Initially I built ICU from source, but in the end I found the Make/Docker-based workflow to be most effective. I really appreciate the work to make this so turn-key. I'm new to the Rust ecosystem, so any pains in getting up and running were largely due to my lack of familiarity with Cargo, bindgen, etc.

I relied heavily on the existing code and documentation. There is some minor variance among some of the crates with regard to APIs, so I didn't have complete confidence in my API design decisions. In most cases, the Rust wrapper type mirrors the name of the underlying ICU type, but `rust_icu_uenum::Enumeration` and `rust_icu_ecma402::PluralRules` are exceptions. Another case is `::new` vs. `::try_new` (I'm not familiar enough to know which, if either, is more Rust-ful; I picked `::new`). The most common case is choosing whether to expect a Rust type or an ICU wrapper type as an argument, e.g., `&str` or `&rust_icu_ustring::UChar`. Ultimately I added `_ustring` variants of such APIs (`rust_icu_unum` was a good example of this precedent).<sup>1</sup> There is also some variance in the amount of documentation. Once I figured out the difference between `//`, `///`, and `//!`, I erred on the side of more documentation than less for the bindings I added. I fully realize that retrofitting C APIs can be awkward, so I hope that my comments are received less as criticism than as a commentary on the difficulty of achieving "consistency in awkwardness."

I'm familiar with PyICU for Python, and having looked over the state of affairs in Rust, I was really glad to discover this project. I'm selfishly contributing bindings in the order that I need them, but I would be happy to continue contributing beyond that. One thing I'm curious about is what your vision is for this project. Is the goal to achieve 100% coverage? How are you prioritizing what gets implemented?

<sup>1</sup> Highlighting this comment from the README

> A low level ICU API may not be an appropriate seam for the end users. A rust-ful API should be layered on top of these bindings.

I can certainly appreciate this in light of some the ICU APIs (for example, [`ubrk_countAvailable`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ubrk_8h.html#a8b587f3b0df1b680f4ca3f1d1f151f7e) and [`ubrk_getAvailable`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ubrk_8h.html#a12fc845896682221749f03e7dc21e3e0) for fetching the list of locales with text breaking info). To me this gets a little blurry when it comes to offering APIs for Rust types in addition to the ICU wrapper types.

bdragon:
One thing I forgot to mention: I have noticed that when I run `make static-bindgen` to pull in new bindings, the generated code isn't formatted for some reason. It's trivial to manually format the files, but just thought I'd mention it.

filmil:
> One thing I forgot to mention: I have noticed that when I run `make static-bindgen` to pull in new bindings, the generated code isn't formatted for some reason. It's trivial to manually format the files, but just thought I'd mention it.

Hm, I am not sure what you mean.  Looking at for example https://github.com/google/rust_icu/pull/161/files#diff-b0e76253c324ca4bf4ce8412c0d26a2f (file lib_67.rs), the code seems formatted to me.  What am I missing?  Have you had to
run `rustfmt` manually?

I'm going to make an assumption about what you are actually observing, so apologies if I get it wrong. Normally, unless told otherwise, `bindgen` will subprocess a rust formatter on its output once generated, so the expected result is to have formatted generated code without manual intervention.  

Perhaps the issue is that `make static-bindgen` docker invocation isn't fully hermetic (see https://github.com/google/rust_icu/blob/master/Makefile#L86). It tries to use the rust formatter local to your machine, but fails to find it and then just skips the formatting step.  If so, this would be unexpected to me, since the container used for bindgen ought to have `rustfmt` available.  But perhaps I missed it. If so, it is a bug, and please file a new issue for me to fix it.

A way this could happen is if I have `rustfmt` installed, but the container does not.  That way everything would work just fine for me, but those who don't have `rustfmt` installed would see an issue.  If you want to corroborate, make sure you have `rustfmt` installed in your `$HOME/.cargo/bin` and retry `make static-bindgen`.  I would suppose that you *don't* have `rustfmt` installed, and that the `rust_icu_testenv` container is missing its `rustfmt`. (Install instructions are here: https://github.com/rust-lang/rustfmt#on-the-stable-toolchain)

HTH.

filmil:
Please allow me some time before I get to review this pull request. I will let you know once I start looking at it.  I also owe you answers to the questions you raised above.

filmil:
> Filip Filmar Thank you for the fast response and code review!

Happy to oblige.

> The README made it easy to get started [...]

Thank you for the feedback.  We were going for lowering the barrier to entry, and I am glad to have a data point that seems to corroborate that we have in fact lowered the barrier to entry.

> name of the underlying ICU type, but `rust_icu_uenum::Enumeration` and `rust_icu_ecma402::PluralRules` are exceptions. 

These two are exceptions, but for different respective reasons. `Enumeration` was one of the earliest wrapped types and I admit to having still been exploring the "right" way to set things up.  The whole ECMA402 set of crates, including `ecma402_traits` and `rust_icu_ecma402` are in effect a separate project, set apart from the "rust_icu proper", and which lives together with rust_icu because it was economical to bootstrap it that way.

> Another case is `::new` vs. `::try_new` (I'm not familiar enough to know which, if either, is more Rust-ful; I picked `::new`). 

`try_new` is a bit of a unique situation for ICU: many of the object constructors in ICU may return an error, so a fallible constructor was needed, which `new` typically is not.  `try_from` may have been an appropriate familiar name, but it is part of the `TryFrom` trait and assumes only one input parameter, a template which most of the ICU constructor functions don't fit.

Hence, `try_new`.

> The most common case is choosing whether to expect a Rust type or an ICU wrapper type as an argument, e.g., `&str` or `&rust_icu_ustring::UChar`. Ultimately I added `_ustring` variants of such APIs (`rust_icu_unum` was a good example of this precedent).1 There is also some variance in the amount of documentation. 

The ICU APIs aren't exactly uniform in this way in an obvious manner.  Sometimes the APIs take or return `const char*`, with or without size, sometimes take for or return `UChar` with or without size, without an order or method that I could infer.

My preference is to deal with Rust types as much as possible.  This works well in the following cases:

* ICU takes a pointer+size as parameters, since this is directly compatible with `&str`

Dealing with `UChar` is necessary when

* The API requires it, so we have no choice.
* The API returns indexes into `UChar`, so converting this to `String` would make the indexing information unusable
* You want to chain multiple ICU API calls, each of which does part of the Unicode processing.  For this reason I settled to have both `foo` and `foo_ustring` where applicable, and most of the time this API duality can be automated by macros.

> Once I figured out the difference between `//`, `///`, and `//!`, I erred on the side of more documentation than less for the bindings I added. I fully realize that retrofitting C APIs can be awkward, so I hope that my comments are received less as criticism than as a commentary on the difficulty of achieving "consistency in awkwardness."

Thank you, I am not aware of a better approach here.

> One thing I'm curious about is what your vision is for this project. Is the goal to achieve 100% coverage? How are you prioritizing what gets implemented?

Thanks for asking, let me try to articulate the origins of the project first.

The original reason for this project was to provide ICU-backed Unicode functionality for Fuchsia, because at the time we started, there was none, and we wanted to use more Rust because of its safety properties.  Also, use of ICU is pervasive on Fuchsia, so this functionality would come practically for free in terms of disk and memory use. For example, if you run a browser, you already have ICU on disk and in memory somewhere.  

This makes an ICU wrapper particularly attractive for use in Fuchsia, when compared to custom and native-Rust Unicode solutions. 

Note, there is ICU4X which is a native take on Unicode in Rust which is catching wind (https://github.com/unicode-org/icu4x), while for example the `unic` crates have been somewhat stagnating, so there is a vacuum in which `rust_icu` fits nicely as a polyfill; and could potentially live on forever as a way to use ICU specifically in Rust (because why not!).

The feature set we want the most is client-side support, so the current focus is to cover ECMA402 API with (1) rust_icu, (2) icu4x and (3) a compatibility layer between (1) and (2).  See here: https://github.com/google/rust_icu/projects for details.

As for the rest, the goal is not necessarily to cover 100% of the API surface, as not all of the API surface is of equal interest. I believe that for non-ECMA402 functionality, things of interest are those for which (1) no adequate support is available in Rust otherwise, and (2) there is a general interest in the community.  An example functionality is word breaking and grapheme cluster iteration

> 1 Highlighting this comment from the README
> 
> > A low level ICU API may not be an appropriate seam for the end users. A rust-ful API should be layered on top of these bindings.
> 
> I can certainly appreciate this in light of some the ICU APIs (for example, [`ubrk_countAvailable`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ubrk_8h.html#a8b587f3b0df1b680f4ca3f1d1f151f7e) and [`ubrk_getAvailable`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ubrk_8h.html#a12fc845896682221749f03e7dc21e3e0) for fetching the list of locales with text breaking info). To me this gets a little blurry when it comes to offering APIs for Rust types in addition to the ICU wrapper types.

In this particular case, I believe there is also an `Enumeration` version of the same API exists (or, at least, *other* parts of the library that have this interface also have an iterated interface), which seems to work well in Rust.  Ostensibly it could cause performance issues to iterate always, versus addressing directly, but without (1) proper benchmarks I wouldn't worry about that too much and (2) the API requires you to know which entry to access anyways; so it's iterated to begin with, and I don't think there is performance to be lost.

In general, I think that trying to solve two issues at the same time is perhaps too much of an effort: I think that `rust_icu` should aim to expose the functionality to Rust first, and have a nice API second.  This is because if you are not satisfied with the API surface, you can use safe Rust to make a nicer one yourself. In fact we're already doing so in that ECMA402 project you discovered.

bdragon:
Re: the formatting issue in my environment:

Strangely this is what I was seeing on my machine:

```
$ which rustfmt
~/.cargo/bin/rustfmt
$ rustfmt --version
This version of rustfmt is deprecated. Use rustfmt-nightly. Use --force to run deprecated rustfmt.

usage: rustfmt [options] <file>...
```

I wiped and re-installed `rustup`. Here's what I'm seeing now:

```
$ rustup default nightly
$ rustup toolchain list
stable-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default) (override)
$ rustup component add rustfmt --toolchain nightly
$ cargo +nightly install bindgen
$ which rustfmt
~/.cargo/bin/rustfmt
$ rustfmt --version
rustfmt 1.4.20-nightly (48f6c32 2020-08-09)
$ which bindgen
~/.cargo/bin/bindgen
$ bindgen --version
bindgen 0.55.1
$ make test
$ make static-bindgen
...
The generator requires bindgen to be in PATH; see README.md
make: *** [Makefile:83: static-bindgen-63] Error 1
```

I verified that `bindgen` is in PATH in the container, but `bindgen --version` is returning a non-zero status, hence the error message. Further debugging in the container reveals:

```
/usr/local/cargo/bin/bindgen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /usr/local/cargo/bin/bindgen) 
```

I'll spend some more time investigating. I'm not yet convinced that this isn't unique to my setup.

filmil:
That last error is telling, and I've seen it elsewhere recently.  

I think this means the libc in the container is too old for the bindgen binary that the container picks up from your disk.  A resolution would be to ensure that there is a working bindgen in the container, and not use the system bindgen.  

In the past, I've found this somewhat difficult to achieve, so perhaps some experimentation is needed.

A band-aid would be to update the container, but that will only shift the errors to someone else with a different setup.

bdragon:
@filmil Thank you so much for the thorough responses. This context is extremely helpful.

I have been working through #166 and had been wavering on aspects of the API. This point in particular is a key take-away for me:

> I think that rust_icu should aim to expose the functionality to Rust first, and have a nice API second.

I expect to have those bindings wrapped up next week.

I am excited to learn about icu4x. I will keep a close eye on it, and perhaps see if the owners are open to outside contributors.

Issue #162, opened 2020-08-27
Implement the ECMA402 collator API surface


filmil:

filmil:
Part of https://github.com/google/rust_icu/issues/134

PR #163, opened 2020-08-27
Implements ECMA402 Collator traits

filmil:
This is a common trait implementation for the ECMA402 `Intl.Collator`
object.

Fixes: #162

### DIFF
--- a/ecma402_traits/src/collator.rs
+++ b/ecma402_traits/src/collator.rs
@@ -55,7 +55,7 @@ pub mod options {
     pub enum Punctuation {
         /// Ignore punctuation.
         Ignore,
-        /// Honor punctyation.
+        /// Honor punctuation.
         Honor,
     }
 
@@ -106,9 +106,9 @@ pub struct Options {
 
 impl Default for Options {
     /// Gets the default values of [Options] if omitted at setup.  The
-    /// default values are prescribed in by the [TC39 report][tc39].
+    /// default values are prescribed by the TC39.  See [Collator][tc39c].
     ///
-    /// [tc39]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator
+    /// [tc39c]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator
     fn default() -> Self {
         Options {
             usage: options::Usage::Sort,
--- a/ecma402_traits/src/collator.rs
+++ b/ecma402_traits/src/collator.rs
@@ -1,0 +1,144 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


filmil:
@zbraniecki PTAL?
(Sorry, for some reason I can not add you as reviewer)

ecma402_traits/src/collator.rs
```diff
@@ -0,0 +1,144 @@
...
+        /// Examples: a ≠ b, a ≠ á, a ≠ A.
+        Variant,
+    }
+
+    /// Whether punctuation should be ignored.
+    #[derive(Eq, PartialEq, Debug, Clone)]
+    pub enum Punctuation {
+        /// Ignore punctuation.
+        Ignore,
+        /// Honor punctyation.
```
kpozin, line 58: Typo

ecma402_traits/src/collator.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+/// The options set by the user at construction time.  See discussion at the top level
+/// about the name choice.  Provides as a "bag of options" since we don't expect any
+/// implementations to be attached to this struct.
+///
+/// The default values of all the options are prescribed in by the [TC39 report][tc39lf].
+///
+/// [tc39lf] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator
```
kpozin, line 97: Wrong URL?

ecma402_traits/src/collator.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub punctuation: options::Punctuation,
+    pub numeric: options::Numeric,
+    pub case_first: options::CaseFirst,
+}
+
+impl Default for Options {
+    /// Gets the default values of [Options] if omitted at setup.  The
+    /// default values are prescribed in by the [TC39 report][tc39].
+    ///
+    /// [tc39]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator
```
kpozin, line 111: Wrong URL?

ecma402_traits/src/collator.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct Options {
+    pub usage: options::Usage,
+    pub sensitivity: options::Sensitivity,
+    pub punctuation: options::Punctuation,
+    pub numeric: options::Numeric,
+    pub case_first: options::CaseFirst,
+}
+
+impl Default for Options {
```
kpozin, line 107: If the spec indicates default values for all the enums, we should probably `impl Default` for them as well.

ecma402_traits/src/collator.rs
filmil: Done.

ecma402_traits/src/collator.rs
filmil: I thought about this as well, and concluded that the default value of an enum at the site of use is not necessarily the "global" default.  From that perspective it didn't seem appropriate to prescribe a global default.

PR #165, opened 2020-08-28
Implements the ECMA402 trait for Intl.DateTimeFormat.

filmil:
Closes: #164

### DIFF
--- a/ecma402_traits/src/datetimeformat.rs
+++ b/ecma402_traits/src/datetimeformat.rs
@@ -1,0 +1,275 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Contains the API configuration as prescribed by ECMA 402.
+///
+/// The meaning of the options is the same as in the similarly named
+/// options in the JS version.
+///
+/// See [DateTimeFormatOptions] for the contents of the options.  See the [DateTimeFormat::try_new]
+/// for the use of the options.
+pub mod options {
+    /// The date and time formatting options.
+    #[derive(Eq, PartialEq, Debug, Clone)]
+    pub enum Style {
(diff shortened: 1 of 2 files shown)


filmil:
@zbraniecki PTAL

I am still not sure about the options -- perhaps a better choice can be made.

ecma402_traits/src/datetimeformat.rs
```diff
@@ -0,0 +1,275 @@
...
+
+#[derive(Eq, PartialEq, Debug, Clone)]
+pub struct DateTimeFormatOptions {
+    /// The formatting style to use for formatting the date part.
+    pub date_style: options::Style,
+    /// The formatting style to use for formatting the time part.
+    pub time_style: options::Style,
+    /// The number of fractional seconds to apply when calling `format`.
+    /// Valid values are 0 to 3.
+    pub fractional_second_digits: u8,
```
kpozin, line 188: Should this be strongly typed to enforce valid values?
(Would be nice if Rust had [ranged integers](https://github.com/rust-lang/rfcs/issues/671).)

zbraniecki:
In my ICU4X attempt I did the following logic for https://github.com/zbraniecki/icu4x/blob/datetime/components/datetime/src/options/components.rs and https://github.com/zbraniecki/icu4x/blob/datetime/components/datetime/src/options/style.rs

which uses https://github.com/zbraniecki/icu4x/blob/datetime/components/datetime/src/options/mod.rs

Would it be something worth considering to align here (or there)?

filmil:
> Would it be something worth considering to align here (or there)?

*tl;dr: I'd attempt to do (1) from below, unless you think a better approach is available.*

Ideally, an interface definition should dictate what the implementation has to conform to.  So if at all possible, I'd prefer to try to modify the implementation in icu4x to conform.

In practice, of course, we extract or retrofit interfaces.  

I hope that it would be possible to:

1. Try to have the trait definition dictate the implementation if possible. I did that in `rust_icu`, sometimes at the expense of having to write glue code.
2. Try to keep the trait definition clean of implementation as much as possible.

Now, I *think* we can't go around implementing the options enum, since the configuration is part of the API surface, unless we're willing to have the trait be "stringly" typed. But let me know if I'm mistaken.

It seemed like a reasonable tradeoff to allow for `impl`s of `Options` enum to continue existing here, as they are part of the API surface and all spec-conforming implementations must have them anyways.

zbraniecki:
I'm mostly talking about "nesting" of options into `Style` vs `ComponentsBag`. Do you think it may be a good option here?

ecma402_traits/src/datetimeformat.rs
filmil: Perhaps, but it's not clear to me how a trait would be doing this.  I'd leave it to the implementor of `new`.

PR #170, opened 2020-09-14
Implement ubrk.h

### LINKED ISSUE #166: Implement ubrk.h
Add support for boundary analysis (character, word, line-break, and sentence) as defined in `ubrk.h`.

bdragon:
Adds bindings for [`ubrk.h`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ubrk_8h.html)

Closes #166

## API

| ICU4C API | Rust API |
|---------|----------|
| `ubrk_close` | `impl Drop` |
| `ubrk_next` | `impl Iterator` |
| `ubrk_countAvailable` | `count_available_locales` |
| `ubrk_getAvailable` | `get_available_locale_at` |;
| `ubrk_open` | `try_new`, `try_new_ustring` |
| `ubrk_openRules` | `try_new_rules`, `try_new_rules_ustring` |
| `ubrk_openBinaryRules` | `try_new_binary_rules`, `try_new_binary_rules_ustring` |
| `ubrk_getBinaryRules` | `get_binary_rules` |
| `ubrk_safeClone` | `safe_clone` |
| `ubrk_setText` | `set_text`, `set_text_ustring` |
| `ubrk_setUText` | Not implemented.<sup>1</sup> |
| `ubrk_refreshUText` | Not implemented.<sup>1</sup> |
| `ubrk_previous` | `previous` |
| `ubrk_current` | `current` |
| `ubrk_preceding` | `preceding` |
| `ubrk_following` | `following` |
| `ubrk_first` | `first` |
| `ubrk_last` | `last_boundary`<sup>2</sup> |
| `ubrk_isBoundary` | `is_boundary` |
| `ubrk_getLocaleByType` | `get_locale_by_type` |
| `ubrk_getRuleStatus` | `get_rule_status` |
| `ubrk_getRuleStatusVec` | `get_rule_status_vec` |

---

<sup>1</sup> In light of #6 and the existing UText implementation, it made sense to leave these out for now.

<sup>2</sup> The Rust `Iterator` trait already provides a `last()` method, so I had to name this differently in order to avoid a conflict.


bdragon:
Minor thing: I noticed that when the `simple_drop_impl!` macro is used, `make cov` does not treat the referenced ICU `*_close` API as implemented in the coverage report, so I added an `Implements ...` doc comment above the macro. This caused `make cov` to consider the ICU API implemented, at the expense of a warning:

```
95 | /// Implements `ubrk_close`.
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ rustdoc does not generate documentation for macro invocations
```

filmil:
> Minor thing: I noticed that when the `simple_drop_impl!` macro is used, `make cov` does not treat the referenced ICU `*_close` API as implemented in the coverage report, so I added an `Implements ...` doc comment above the macro. This caused `make cov` to consider the ICU API implemented, at the expense of a warning:
> 
> ```
> 95 | /// Implements `ubrk_close`.
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ rustdoc does not generate documentation for macro invocations
> ```

You will note that elsewhere I use `// Implements ...` instead of `...` where the macro is invoked to work around this issue.
Since the macro expansion itself contains the three-slashed version of the same text, the end result comes out OK in both rustdoc and compilation - without warnings.

bdragon:
Thanks. Updated in d1876d5.

filmil:
I'm now looking at this PR.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+        }
+    }
+}
+
+impl UBreakIterator {
+    /// Reports the number of locales for which text breaking information is
+    /// available.
+    ///
+    /// Implements `ubrk_countAvailable`.
+    pub fn count_available_locales() -> i32 {
```
filmil, line 136: I would replace this and the following methods with an iterator.
This avoids passing an index that is larger than the number of available locales.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+        unsafe { versioned_function!(ubrk_countAvailable)() }
+    }
+
+    /// Returns the locale for which line breaking information is available
+    /// at the specified index.
+    ///
+    /// Implements `ubrk_getAvailable`.
+    pub fn get_available_locale_at(
```
filmil, line 144: As noted above, I recommend not implementing this method, but rather implementing an iterator over available locales.

Here it's possible to call with `index  >= count_available_locales()`, which should either be guarded with an assertion, or avoided altogether.

bdragon:
I found a couple of broken links in the doc comments. Fixed in 1932074.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+    // original iterator, while maintaining its own internal iteration state.
+    //
+    // For these reasons, these fields are wrapped in `Rc`, ensuring that the
+    // referenced values (`text`, `locale`, etc.) are not released prematurely
+    // if the original break iterator is dropped before its clone. As break
+    // iterators are inherently not thread-safe [1], `Rc` was chosen over `Arc`.
+    //
+    // [1] https://unicode-org.github.io/icu/userguide/boundaryanalysis/#thread-safety
+    text: Rc<ustring::UChar>,
+    locale: Option<Rc<ffi::CString>>,
```
filmil, line 90: Note, you could use `ULoc` here instead of `ffi::CString` in case that helps handling the underlying C string.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+            iter.get_locale_by_type(ULOC_ACTUAL_LOCALE).unwrap(),
+            "root"
+        );
+    }
+
+    #[test]
+    fn test_available_locales() {
+        trace!("Available locales");
+        let count = UBreakIterator::count_available_locales();
+        for i in 0..count {
```
filmil, line 727: what happens if `i > count`?

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,735 @@
...
+    use super::UBreakIterator;
+    use log::trace;
+    use rust_icu_sys::{UBreakIteratorType::*, ULocDataLocaleType::*};
+    use std::{convert::TryFrom, rc::Rc};
+
+    const TEXT: &str =
+        r#""It wasn't the wine," murmured Mr. Snodgrass. "It was the salmon.""#;
+
+    const WORD_BOUNDARIES: [i32; 30] = [
+        0, 1, 3, 4, 10, 11, 14, 15, 19, 20, 21, 22, 30, 31, 33, 34, 35, 44, 45,
```
filmil, line 545: It is a bit hard to check that the result is correct when just numeric.  Could you give a test case that also gives an example of using the returned breaks to extract text fragments?  What would it take to convert the word boundaries into an iterator of `&str`for example, that shows how `TEXT` was split?

Makefile
```diff
@@ -121,6 +121,7 @@ endef
 .PHONY: publish
 publish:
 	$(call publish,rust_icu_sys)
+	$(call publish,rust_icu_ubrk)
```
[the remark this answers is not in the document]
filmil: You may also want to weave it into the `rust_icu` crate.https://docs.rs/rust_icu/0.3.2/rust_icu/

filmil (changes requested):

Makefile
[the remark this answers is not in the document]
bdragon: Thanks for catching. Addressed in 3f9d961 and 3167821.

rust_icu_ubrk/src/lib.rs
bdragon: My intuition was to wrap these up as an iterator, so I feel somewhat validated to learn that you would have done the same. I think that I may have been a bit too literal-minded when reading the part of your [comment](https://github.com/google/rust_icu/pull/161#issuecomment-686054854) on my last PR related to leaving some aspects of API refinement to end users. Refactored in 70f4642.

rust_icu_ubrk/src/lib.rs
bdragon: Great. I hadn't thought of that. Addressed as part of 70f4642.

rust_icu_ubrk/src/lib.rs
bdragon: This is moot following the change to replace these methods with an iterator, but I was curious. It turns out the underlying API will happily return a NULL pointer, and thus `get_available_locale_at` will return None. Safe, but inelegant.

rust_icu_ubrk/src/lib.rs
bdragon: Great. I hadn't thought of that. Addressed as part of 70f4642.

rust_icu_ubrk/src/lib.rs
bdragon: Good idea. I think iterating over `&str` also aligns more closely with a typical use case for a break iterator. I will address this next.

bdragon: I tried to address this in cea1b59. Please let me know if it doesn't align with what you had in mind.

bdragon:
@filmil I appreciate the review. I believe I have addressed your feedback if you don't mind taking another look.

rust_icu_ubrk/src/lib.rs
bdragon: Correction: I added an iterator that exposes available locales as `ULoc` in 70f4642.

I have just now updated `try_new_ustring` to accept locale as a `ULoc` (2fce06f), which I think makes more for a better API, but ultimately the value that is owned by the iterator is the `CString`, so I did not change the type of the field in the struct.

filmil:
Thank you for making the changes, I'll take a look as soon as able, but at the moment it seems will need to happen tomorrow.

filmil:
Re-reviewing now.

rust_icu_ubrk/src/lib.rs
filmil: OK.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+            rules: Some(Rc::new(rules)),
+            binary_rules: None,
+            text: Rc::new(text),
+            rep: ptr::NonNull::new(rep).unwrap(),
+        })
+    }
+
+    /// Creates a new break iterator using pre-compiled binary rules.
+    ///
+    /// Binary rules can be obtained with [`get_binary_rules`].
```
filmil, line 252: I think that if you use the qualified name of `get_binary_rules`, it will automatically get linked and formatted with fixed spacing font.

Like so:

```
[UBreakIterator::get_binary_rules]
```

No need to change now, but it is something to be aware of and can make documenting a bit easier.

rust_icu_ubrk/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+
+        let text_len = TEXT.len() as i32;
+        assert!(break_iter.is_boundary(text_len));
+        assert_eq!(text_len, break_iter.last_boundary());
+        assert_eq!(None, break_iter.next());
+        assert_eq!(text_len, break_iter.current());
+
+        break_iter.first();
+        let word_iter = Words::new(&mut break_iter);
+        assert_eq!(
```
filmil, line 648: Thanks for making this change.  Is there a specific reason that `word_iter` takes `break_iter`as mutable?

filmil (approved):

filmil:
Thank you for your patience for the review.  This is now approved.

rust_icu_ubrk/src/lib.rs
bdragon: Calling `next()` on an iterator requires a mutable reference, and since `Words` calls `next()` on the break iterator, it needs to hold a mutable reference. Alternatively it could fully take ownership of the break iterator, since this is just a helper type for tests.

PR #172, opened 2020-09-18
Makes ecma402_traits an optional dependency

### LINKED ISSUE #171: Make the dependency on ecma402_traits in `rust_icu_uloc` conditional on a feature
There is a marker implementation of `ecma402_traits::Locale` for `ULoc` here: https://github.com/google/rust_icu/blob/master/rust_icu_uloc/src/lib.rs#L58

This marker is not useful if you don't want to use any ecma402 traits, so it should probably be hidden behind a feature.

filmil:
This removes the need for every use of `rust_icu_uloc` to
depend on `ecma420_traits` if the marker implementation of
`ecma402_traits::Locale` is not in fact being used.

This happens for example when you are using `rust_icu`
directly and not through any ECMA402 compatible API. This change
should, for example, remove the need for Fuchsia code base to
vendor ecma402_traits.

Closes #171.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         icu_version: [63]
         feature_set: 
-          - "renaming,icu_version_in_env"
+          - "renaming,icu_version_in_env,use_ecma402"
     steps:
       - uses: actions/checkout@v2
       - name: Test ICU version ${{ matrix.icu_version }}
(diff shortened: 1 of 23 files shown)


kpozin:
Is it worth adding a blurb to the "Features" sections of README.md, or is this still considered too experimental?

kpozin (approved):

filmil:
> Is it worth adding a blurb to the "Features" sections of README.md, or is this still considered too experimental?

Added, no reason to keep it a secret.

Issue #175, opened 2020-09-28
Add support for ICU 68


filmil:
The official release of ICU 68 is nearing.  We should ensure that test fixtures are present to test ICU68.

filmil:
Note, since ICU versions as old as 63 are still being shipped with some Linux distributions for example, we can't really drop support for very old versions.

We keep feature flags that allow us to support bugfixes and features of new libraries.

filmil:
ICU 68.1 is now released, I'm taking the upgrade.

filmil:
List of things to do:

* Introduce a "68_plus" feature flag (since icu data varies)
* Add the "68" buildenv
* Add `lib68.rs`
* Add continuous testing for 68.

filmil:
This is now fixed, 0.4.1 has support for ICU 68.1

PR #178, opened 2020-10-28
Upgrades the library to support ICU 68.

filmil:
Issue #175


build/Makefile
```diff
@@ -46,6 +46,8 @@ all: \
 	push-testenv-66 \
 	push-maint-67 \
 	push-testenv-67 \
+	push-maint-67 \
+	push-testenv-67 \
```
kpozin, line 50: These should be `68`, right?

filmil: Done.

kpozin (approved):

Issue #177, opened 2020-10-28
Re-enable Arabic tests in udat.rs


filmil:
I had to disable Arabic tests in udat.rs as the feature flag does not seem to work when using `make docker-test`, and I'm not sure exactly why.

Using `make test` i.e. testing with the system ICU, works as expected.

PR #180, opened 2020-10-28
Fix tests icu 68.1

filmil:
Issue #177


kpozin (approved):
Looks good, but the commit description seems to be from a different commit.

Issue #189, opened 2020-12-10
Building rust_icu on NixOS


suhr:
The compilation of `rust_icu` on NixOS fails with the following message:

```
error: failed to run custom build command for `rust_icu_sys v0.4.1 (registry `ssh://git@gitlab.radicum.team/common-utils/crates-index.git`)`

Caused by:
  process didn't exit successfully: `/home/suhr/Projects/work/db/local-target/release/build/rust_icu_sys-11ae63ac84826367/build-script-build` (exit code: 1)
  --- stdout
  cargo:rustc-cfg=feature="renaming"
  cargo:rustc-cfg=feature="use-bindgen"
  cargo:rustc-cfg=feature="icu_config"
  icu-version-major: 67
  cargo:rustc-cfg=feature="icu_version_64_plus"
  cargo:rustc-cfg=feature="icu_version_67_plus"
  icu-version: 67.1
  icu-cppflags: -I/nix/store/7flw8mxs28mk2r1wsnyas1y59j4hjr7a-icu4c-67.1-dev/include
  icu-has-renaming: true
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/ubrk.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/ucal.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/uclean.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/ucnv.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/ucol.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/udat.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/udata.h'
  include-file: '/nix/store/wn2hcvdk7dzc7l3f340sva6lhkbmmsgd-icu4c-67.1/include/unicode/uenum.h'
[...]

PR #192, opened 2021-01-12
[rust_icu_sys] Gets build.rs out of the business of caring about incl…

### LINKED ISSUE #191: #190 can break rust_icu_sys build
The recent pull request #190 breaks the build of `rust_icu_sys` on my Fedora 33 machine.

On my computer, `pkg-config --cflags icu-i18n` prints out a blank line, because the only include directory is the default system search path `/usr/include`.

This causes a problem here:

https://github.com/google/rust_icu/blob/2d5cfd806fa4dedd96abe861b960760773e1564b/rust_icu_sys/build.rs#L267

The loop doesn't execute, so only `f` is passed to the next stage of the expression, resulting in lines that look like:

```
#include "ubrk"
```

which causes `bindgen` to fail.

It looks like this could be avoided by passing `--keep-system-cflags` to pkg-config. I don't know how portable that is?

filmil:
…udes

In #190, I added a check for specific include files that are added to
the include generated include file by build.rs.  Sadly this does not
work in the case where `--cppflags` is empty, i.e. in the case of
systems where ICU is installed in say `/usr/include`.

In this second attempt, `build.rs` gets out of the business of guessing
the include paths, but instead only passes the `-I` flag to bindgen,
while providing a relative path to the relevant include files, which
is always of the form `#include "unicode/foo.h"`.

An alternative of using the option `--keep-system-cflags` may end up
needing even more fixups in `build.rs` to work reliably.  So we're not
doing that.

Closes #191.

### DIFF
--- a/rust_icu_sys/build.rs
+++ b/rust_icu_sys/build.rs
@@ -207,15 +207,6 @@ mod inner {
             self.prefix()
         }
 
-        /// Returns all the include directories as reported by the C++ flags.
-        fn include_dirs(&mut self) -> Result<Vec<String>> {
-            Ok(self.cppflags()?.split(" ")
-                .filter(|p| p.starts_with("-I"))
-                .map(|p| &p[2..])
-                .map(|s| s.to_string())
-                .collect())
-        }
-
         /// Returns the config major number.  For example, will return "64" for
         /// version "64.2"
         fn version_major() -> Result<String> {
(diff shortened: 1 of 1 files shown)


filmil:
@pkgw @suhr would y'all mind trying out this pull request to check that:

- issue #192 is fixed (for @pkgw)
- issue #189 didn't regress (for @suhr)

Asking you since we currently don't have either Fedora or a NixOS CI setups. Thanks!

pkgw:
Confirmed that it fixes the build for my system.

filmil:
Thank you. I'll leave this unmerged for a little while longer to see if @suhr can corroborate.

kpozin (approved):

pkgw:
For after this gets merged ... what's the release schedule looking like? I can't use the most recent release because I build on stable, as well as this issue on my personal machine.

filmil:
For the time being, the releases are on an as-needed basis. But it is no
issue to release a patch level - please file a bug so I don't forget.
Thanks.

suhr:
Sorry for delay. I'll test this a bit later.

filmil:
I'll need to make some repo changes, so let's merge this.  Worst case, it fixes one, and breaks another build, so net zero.

PR #195, opened 2021-02-27
Changes in preparation for implementing Intl.DateTimeFormat.

filmil:
I realized that, in order to initialize the date-time formatter
propertly, I will need to admit changes to a once-defined locale.

So, adding the code that does that.

See: #135

### DIFF
--- a/rust_icu_ecma402/src/datetimeformat.rs
+++ b/rust_icu_ecma402/src/datetimeformat.rs
@@ -33,6 +33,7 @@ pub(crate) mod internal {
     use rust_icu_ustring as ustring;
     use std::convert::TryFrom;
 
+    // TODO: implement this conversion completely.
     pub fn opt_to_pattern(_opts: DateTimeFormatOptions) -> Result<ustring::UChar, common::Error> {
         ustring::UChar::try_from("YYYY")
     }
--- a/rust_icu_ecma402/src/datetimeformat.rs
+++ b/rust_icu_ecma402/src/datetimeformat.rs
@@ -1,0 +1,126 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implements the traits found in [ecma402_traits::datetimeformat].
+
+use ecma402_traits;
+use rust_icu_common as common;
+use rust_icu_udat as udat;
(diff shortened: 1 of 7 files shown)


rust_icu_uloc/src/lib.rs
```diff
@@ -562,6 +674,19 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_keywords_nounicode() -> Result<(), Error> {
+        let loc = ULoc::for_language_tag("az-Cyrl-AZ-u-ca-hebrew-t-it-x-whatever")?;
```
kpozin, line 679: For completeness, please add key-value pairs to one of the other extensions (e.g. `x-key-whatever-value`).

rust_icu_uloc/src/lib.rs
```diff
@@ -31,6 +32,109 @@ use {
 /// See `ULOC_FULLNAME_CAPACITY`.
 const LOCALE_CAPACITY: usize = 158;
 
+/// [ULocMut] is a mutable companion to [ULoc].
```
kpozin, line 35: Would `ULocBuilder` be an accurate name?

rust_icu_uloc/src/lib.rs
```diff
@@ -763,4 +888,56 @@ mod tests {
...
+
+        let loc = ULoc::for_language_tag("en-t-it-u-tz-uslax-x-foo")?;
+        let mut loc_mut = ULocMut::from(loc);
+        loc_mut.set_unicode_keyvalue("tz", "usnyc");
+        loc_mut.set_unicode_keyvalue("nu", "arabic");
+        let loc = ULoc::from(loc_mut);
+        assert_eq!(ULoc::for_language_tag("en-t-it-u-nu-arabic-tz-usnyc-x-foo")?, loc);
+        assert_eq!(ULoc::for_language_tag("en-t-it-u-tz-usnyc-nu-arabic-x-foo")?, loc);
+        Ok(())
+    }
```
kpozin, line 942: One more test case for completeness: calling `set_unicode_keyvalue()` on a locale that originally only has non-`u` extensions.

kpozin (approved):

rust_icu_ecma402/src/datetimeformat.rs
```diff
@@ -0,0 +1,126 @@
...
+    rep: udat::UDateFormat,
+}
+
+pub(crate) mod internal {
+    use ecma402_traits::datetimeformat::DateTimeFormatOptions;
+    use rust_icu_common as common;
+    use rust_icu_ustring as ustring;
+    use std::convert::TryFrom;
+
+    pub fn opt_to_pattern(_opts: DateTimeFormatOptions) -> Result<ustring::UChar, common::Error> {
```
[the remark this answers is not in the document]
filmil: Done.

rust_icu_uloc/src/lib.rs
filmil: I don't think so, since the collaboration doesn't fit the build model.

rust_icu_uloc/src/lib.rs
filmil: This setting is not implemented at all for the time being.  Once it is, it will make sense to add such a test.

PR #205, opened 2021-05-03
Implements ECMA402 compatible date and time formatting.

filmil:
Adds ECMA 402 compatible options for date and time formatting.  This is
not 100% aligned with ECMA since some options are not quite clear.  But
it is close enough, and can be used out of the box.

Fixes: #135


rust_icu_ecma402/src/datetimeformat.rs
```diff
@@ -29,13 +29,147 @@ pub struct DateTimeFormat {
...
+        }
+        if let Some(second) = &opts.second {
+            match second {
+                options::DisplaySize::Numeric => skel.push("s"),
+                options::DisplaySize::TwoDigit => skel.push("ss"),
+            }
+        }
+        if let Some(time_zone_style) = &opts.time_zone_style {
+            match time_zone_style {
+                options::TimeZoneStyle::Long => skel.push("VV"),
```
kpozin, line 145: Are you sure about time zone styles? According to the syntax guide:

 description | skeleton | output
--- | --- | ---
Time Zone: short time zone ID | V | uslax
Time Zone: long time zone ID | VV | America/Los_Angeles

These are not meant for display to end-users.

Testing `Intl.DateTimeFormat` in the browser, I get

timeZoneName | output | probable skeleton
---|---|---
short | GMT-8 | O
long | Pacific Daylight Time | zzzz

kpozin (changes requested):

ecma402_traits/src/datetimeformat.rs
```diff
@@ -191,7 +204,7 @@ pub struct DateTimeFormatOptions {
     pub time_style: options::Style,
     /// The number of fractional seconds to apply when calling `format`.
     /// Valid values are 0 to 3.
-    pub fractional_second_digits: u8,
+    pub fractional_second_digits: Option<u8>,
```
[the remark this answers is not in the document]
filmil: That's unclear, I'd need to check.

rust_icu_ecma402/src/datetimeformat.rs
filmil: Yeah, I'm not 100% sure as to which skeleton needs to be used. :/

ecma402_traits/src/datetimeformat.rs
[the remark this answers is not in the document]
filmil: Checked: value of 0 results in `fractionalSecondDigits value is out of range`.

rust_icu_ecma402/src/datetimeformat.rs
filmil: hm, `uloc_forLanguageTag` fails for any locale that has a country component, but not if the locale has only language.  For example, `sr` works, but `sr-rs` does not.  Weirdness.

filmil: Yeah, looks like underscores don't work: `en-us` is OK as a language tag but `en_US` isn't.  (Looks like while this has been documented for a while, it only recently started actually being enforced, based on the fact that older ICU builders don't have an issue.)

filmil: Fixed for timeZoneName.

filmil:
PTAL.

ecma402_traits/src/datetimeformat.rs
```diff
@@ -186,12 +201,16 @@ pub mod options {
...
+    /// If `date_style` or `time_style` are set, none of the other options
+    /// are acceptable.
+    pub date_style: Option<options::Style>,
     /// The formatting style to use for formatting the time part.
-    pub time_style: options::Style,
+    /// If `date_style` or `time_style` are set, none of the other options
+    /// are acceptable.
+    pub time_style: Option<options::Style>,
     /// The number of fractional seconds to apply when calling `format`.
     /// Valid values are 0 to 3.
```
kpozin, line 212: Valid values are 1 to 3.

rust_icu_ecma402/src/datetimeformat.rs
```diff
@@ -29,13 +29,179 @@ pub struct DateTimeFormat {
...
+    use rust_icu_uloc as uloc;
     use std::convert::TryFrom;
+    use anyhow::anyhow;
 
-    // TODO: implement this conversion completely.
-    pub fn opt_to_pattern(_opts: DateTimeFormatOptions) -> Result<ustring::UChar, common::Error> {
-        ustring::UChar::try_from("YYYY")
+    fn check_auto_styling_set(opts: &DateTimeFormatOptions) -> Result<(), common::Error> {
+        if opts.date_style.is_some() || opts.time_style.is_some() {
+            return Err(common::Error::Wrapper(anyhow!("date_style or time_style is set")));
```
kpozin, line 41: The error message isn't entirely clear. Maybe something like "can't use auto styling because date_style or time_style is set".

kpozin (approved):
Please run `cargo fmt`.

PR #208, opened 2021-07-23
collation: add sort key support

vmiklos:
By adding a safe wrapper around ucol_getSortKey().

### DIFF
--- a/rust_icu_ucol/src/lib.rs
+++ b/rust_icu_ucol/src/lib.rs
@@ -133,6 +133,35 @@ impl UCollator {
         UCollator::to_rust_ordering(result)
     }
 
+    /// Get a sort key for a string from this collator.
+    ///
+    /// Returns a sort key.
+    ///
+    /// Implements `ucol_getSortKey`
+    pub fn get_sort_key(&self, source: &ustring::UChar) -> Vec<u8> {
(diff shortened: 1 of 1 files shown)


vmiklos:
Thanks! :-) Just curious, do you have any plans when would you cut a next v1.x.y release (from rust_icu_ucol or in general)?

filmil:
New releases are cut on an as needed basis - I'd be glad to make one if you need it, but please allow some time for that to happen.

vmiklos:
Makes sense, no rush. :-) I need this because I'm porting a python prototype to rust and pyicu provides this API. But no problem to just use git master as an interim solution till you're comfortable to push a new tag. Thanks!

vmiklos:
Ping, could you please cut a new release at some stage?

I still have this as a TODO at

https://github.com/vmiklos/osm-gimmisn/blob/224017da7fe9ed46981767f3a0d75d855374b40d/Cargo.toml#L40-L43

Thanks.

filmil:
Oops, I forgot.  I'll try to do it next week, fwiw I have a few other minor improvements lined up.

filmil:
FYI, 1.0.1 is now published.

vmiklos:
Thanks a lot! :-)

PR #213, opened 2021-09-10
Adopts the C++ comments from the ICU library

filmil:
Earlier versions of cargo insisted on running comments generated
from C++ code as rust tests when the comment lines were indented.

This is of course not correct behavior for generated code. Since at
the time there was no way to turn the behavior off, we had to turn
off documentation generation.

Now that the behavior seems to be implemented, we can start
generating docs again.


filmil:
turns out we already do have a section

```
[lib]
doctest = false
```

in `rust_icu_sys/Cargo.toml`.  So, let's start using it.

kpozin (approved):

PR #217, opened 2021-10-09
unorm2: partial support for normalization

filmil:
Just enough to make keyboard dead keys work as required.


rust_icu_unorm2/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Debug)]
+pub struct UNormalizer {
+    rep: std::ptr::NonNull<sys::UNormalizer2>,
+    owned: bool,
+}
+
+impl Drop for UNormalizer {
+    /// Implements `unorm2_close`
+    fn drop(&mut self) {
+        // Close the normalizer only if it is unowned.
```
kpozin, line 33: `only if it is owned`? Or am I misreading that?

rust_icu_unorm2/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+        let result = self.normalize_ustring(&norm)?;
+        String::try_from(&result)
+    }
+
+    /// Implements `unorm2_normalize`.
+    pub fn normalize_ustring(
+        &self,
+        norm: &ustring::UChar
+        ) -> Result<ustring::UChar, common::Error> {
+        const CAPACITY: usize = 200;
```
kpozin, line 92: Could we take a better guess here? We have the string to be normalized, so maybe some multiple of its length?

filmil: We possibly could. However, allowing non-`const` here would require reworking the macro, which I am reluctant to do for unclear gain.

kpozin (approved):

PR #218, opened 2021-10-09
Uchar oct8

filmil:
Adds a partial implementation for uchar.h, the Unicode properties queries and data.


rust_icu_uchar/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    result.into()
+}
+
+/// See <http://www.unicode.org/reports/tr44/#Canonical_Combining_Class_Values> for
+/// the list of combining class values.
+///
+/// Implements `u_getCombiningClass`
+///
+/// Since 1.0.2
+pub fn get_combining_class(c: sys::UChar32) -> u8 {
```
kpozin, line 61: Might be nice to add an enum for these. Up to you.

kpozin (approved):

Issue #223, opened 2021-11-11
Add support for ICU 70.1


filmil:
ICU 70.1 is released.  

Make sure that `rust_icu` works with it.

filmil:
This is going to be tricky.

- I found out that, in order for this update to work, we'll need to update buildenv first.
  - To do that, we need to rebuild the buildenv software
    - To do that, we must update the build image to a more recent rust compiler (else things won't work due to glibc mismatch)
      - To do that, we must resolve https://github.com/kennytm/extprim/issues/22 on a package which seems to be dormant for 4 years now, but is required for rustfmt to work.

filmil:
Looks like most issues will be handled by "just" going back to the stable toolchain.
https://github.com/google/rust_icu/issues/225 handles that.

filmil:
Inching towards 70.1 support.

PR #224, opened 2021-11-11
Icu 70

filmil:
Makes the necessary changes to have rust_icu compile with ICU 70.1
    
The two changes needed were timezone data version regex upgrade and a test touch-up.
    
This has been tested locally. Followup work will provide the builders.
    
Issue: #223

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         icu_version: [63]
         feature_set:
-          - "renaming,icu_version_in_env"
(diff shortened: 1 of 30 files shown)


kpozin (approved):
Nit: For clarity, I would consider changing these feature names to `min_icu_version_68`, `max_icu_version_69`, etc. The juxtaposition of "plus" and "max" suffixes sounds less intuitive.

filmil:
Noted. I'll merge this so that I have a flag to work with, and will follow up with a rename.

Issue #229, opened 2021-11-28
Character set conversion support (`ucnv.h`)


lacc97:
Are there any plans to support the ICU character set conversion API (essentially the `ucnv.h` header)?

filmil:
Beyond release 1.0.0, additions are on an as-needed basis.  Feel free to
contribute the bindings that you need but see are missing.

filmil:
This was, in fact, done.

PR #230, opened 2021-11-30
Implement character encoding conversion (`ucnv.h`)

lacc97:
This PR implements wrapping functions and types on top of ICU's `ucnv.h` header (solves #229).

This PR adds a new crate called `rust_icu_ucnv` in line with the naming convention of the other crates in this project.

The main type is UConverter, which can be created using `UConverter::open` with an encoding name (as a `&str`). This type provides conversion functions between UTF-16 and the provided encoding.

This crate also provides `utf8::Converter` as a convenience type to work directly with UTF-8 strings, such as with Rust’s `&str` and `String` types.


lacc97:
Okay, I have signed the CLA now.

filmil:
You will need to rerun `make static-bindgen` to generate the newly-available functions for the pre-generated bindings.

To do that, please edit `rust_icu_sys/bindgen/run_bindgen.sh`  the same way you did with `build.rs` files - adding the missing header names and regexes.
This will allow the remaining builders to pass.

filmil:
Also, thank you for your patience for the review.

lacc97:
I haven't been able to get the docker environment set up properly. It complains with 
```
env: './rust_icu/rust_icu_sys/bindgen/run_bindgen.sh': Permission denied
```
when I run `make static-bindgen`.

I've only updated the run_bindgen.sh file.

lacc97:
Wait, nevermind. I think I got it working in the end.

PR #240, opened 2022-04-11
Update 71.1

filmil:
- Removes brittle test configurations
- Adds ICU 71.1 to the build and test containers


kpozin (approved):
LGTM, modulo that test that need to be re-run.

filmil:
Thank you. The test failures look like infra issues to me.  I forced a re-run, let's see.

PR #243, opened 2022-04-19
Wrap ucol_setMaxVariable

hsivonen:
The use case is allowing the use of `rust_icu` upstream for fuzzer-based comparison of ICU4X and ICU4C results.


filmil:
Thank you for the contribution!

Issue #244, opened 2022-04-20
normalize_ustring hard-codes the output capacity to 200


hsivonen:
Normalization fails if the output of the operation is longer than 200 UTF-16 code units, because the [output capacity is hard-coded](https://github.com/google/rust_icu/blob/e56a1a3452365f37964053ad16e5b67433cc40ae/rust_icu_unorm2/src/lib.rs#L96). Expected the code to deal with variable-length output.

filmil:
What error are you getting, and could you provide a repro test?

The output capacity you linked to is not hard coded; rather it's the capacity of the initial buffer - the idea is if the initial buffer is not overflowed that we don't need to redo the call. But if it overflows, we repeat the call using the needed buffer size that was discovered during preflight.

See here: https://docs.rs/rust_icu_ustring/latest/src/rust_icu_ustring/lib.rs.html#86

hsivonen:
Here's the repro (67 Hangul syllables that decompose to 3 jamo each, i.e. 201 jamo):

```rust
use rust_icu_ustring::UChar;
use rust_icu_unorm2::UNormalizer;

fn main() {
    let s = "탐탐퀀탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐탐";
    let u = UChar::try_from(s).unwrap();
    let normalizer = UNormalizer::new_nfd().unwrap();
    normalizer.normalize_ustring(&u).unwrap();
}
```

The failure with `rust_icu` 2.0.2 and system ICU from Ubuntu 20.04 is (but I first saw this with `rust_icu` trunk and ICU4C nearly-current trunk):
`
thread 'main' panicked at 'assertion failed: common::Error::is_ok(status)', /home/hsivonen/.cargo/registry/src/github.com-1ecc6299db9ec823/rust_icu_unorm2-2.0.2/src/lib.rs:97:9`

filmil:
> Here's the repro (67 Hangul syllables that decompose to 3 jamo each, i.e. 201 jamo):

Thank you I have a fix.

filmil:
@hsivonen you can now try 2.0.3 and see if it works for you. 

Thank you for your patience.

hsivonen:
Thanks. My test case now passes.

PR #245, opened 2022-04-21
Fix the buffer preflight treatment and add tests

filmil:
The character conversion code uses a two-pass approach for converting
potentially large inputs into various Unicode strings.  The first
pass uses a fixed capacity buffer (often 200 units).  For strings
that happen to be shorter, the conversion succeeds and nothing else
needs to be done.

But for longer strings, the conversion fails, returning a non-OK status
and the size of the buffer that is actually needed.  A second pass then
provides a buffer that's exactly long enough.

That's great and all, but `status` is reused across both calls, and when
the conversion fails, `status` needs to be set back to `OK_CODE` before
the 2nd pass is attempted.

Props for having the assertion that checks for this; but boos for not
actually testing the conversion code with long strings.

Fixes: #244

### DIFF
--- a/rust_icu_uloc/src/lib.rs
+++ b/rust_icu_uloc/src/lib.rs
@@ -585,6 +585,16 @@ mod tests {
         Ok(())
     }
 
+    // https://github.com/google/rust_icu/issues/244
+    #[test]
+    fn test_long_language_tag() -> Result<(), Error> {
+        let mut language_tag: String = "und-CO".to_owned();
+        let language_tag_rest = (0..500).map(|_| " ").collect::<String>();
+        language_tag.push_str(&language_tag_rest);
+        let loc = ULoc::for_language_tag(&language_tag)?;
+        Ok(())
+    }
+
     #[test]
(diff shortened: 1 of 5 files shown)


rust_icu_unorm2/src/lib.rs
```diff
@@ -151,4 +152,16 @@ mod tests {
         }
         Ok(())
     }
+
+    // https://github.com/google/rust_icu/issues/244
+    #[test]
+    fn test_issue_244() -> Result<(), common::Error> {
```
kpozin, line 158: `test_long_input()`?

rust_icu_uloc/src/lib.rs
```diff
@@ -585,6 +585,16 @@ mod tests {
         Ok(())
     }
 
+    // https://github.com/google/rust_icu/issues/244
+    #[test]
+    fn test_issue_244() -> Result<(), Error> {
```
kpozin, line 590: `test_long_language_tag()`?

kpozin (approved):

PR #249, opened 2022-07-01
Fix build.rs for when renaming is off.

Berrysoft:
The macro here mistakenly uses double braces, which make the compiler complain.


filmil:
Thank you!

PR #253, opened 2022-08-01
Implement `buffered_string_method_with_retry` as a function

SabrinaJewson:
It is far simpler if this macro is just a function that accepts a closure.


filmil:
Good idea.

PR #254, opened 2022-08-01
Fix use of metavars in doc comments

SabrinaJewson:
Metavars don’t work in doc comments, as can be seen on docs.rs on [methods like this one](https://docs.rs/rust_icu_ucol/latest/rust_icu_ucol/struct.UCollator.html#method.set_max_variable). This can be fixed by using a manual `#[doc]` attribute with `concat!` and `stringify!`,


filmil:
#til

Issue #228, opened 2021-11-25
Feature for static linking build


izderadicka:
I want to use `rust_icu_col` in building fully statically linked binary.  I'm building on alpine with RUSTFLAGS="-C target-feature=+crt-static -C link-self-contained=yes"

Problem here is that I'm getting this error when rust_icu_sys is compiled:
```
thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at /usr/lib/libclang.so.11.1 could not be opened: Dynamic loading not supported"', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.53.3/src/lib.rs:1956:31
```

I do have static version of libclang - e.g libclang.a on system,  but build is still preferring dynamic version.

In my other projects I' m using feature `static` which will enforce ` println!("cargo:rustc-link-lib=static=ssl");` (here should be clang) for dependant C libraries.

However I'm not sure how to do it for indirect dependencies - e.g. you depend on C library and I depend on your crate. 

I guess same will be for icu C library, but just hit clang first.

filmil:
It looks like a feature flag would work to emit the correct settings. WDYT?

izderadicka:
Yes, feature flag  would be great.

I finally was able to do a static build against rust_icu_col, but it took this:
-  build in alpine with taget x86_64-unknown-linux-musl
-  need to install static libs for icu a clang
- use rust_icu_col without default features, just with `rename` and  `icu_version_in_env`
- use environment vars:
```
export RUST_ICU_MAJOR_VERSION_NUMBER=67
export RUSTFLAGS="-C target-feature=+crt-static -C link-self-contained=yes -L native=/usr/lib -l static=clang -l static=icui18n -l static=icuuc -l static=icudata -l static=stdc++"
```

Some of the complexity is caused by Rust complicated approach to static build, but for rest `static` feature in rust_icu will help.
From what I have seen bindgen has static feature too, which might be useful.

filmil:
Feel free to contribute a patch that handles your use case.  It looks like you've already done the legwork, so I hope it shouldn't be too complicated.

PR #255, opened 2022-08-02
Add `static` feature flag

SabrinaJewson:
This is not the same solution as in #228, but it works perfectly fine for me. Also, note that I’m not too well-versed in linking so I may have made a mistake here.

Closes: #228

### DIFF
--- a/rust_icu/Cargo.toml
+++ b/rust_icu/Cargo.toml
@@ -174,6 +174,7 @@ icu_version_69_max = [
   "rust_icu_utext/icu_version_69_max",
   "rust_icu_utrans/icu_version_69_max",
 ]
+static = ["rust_icu_sys/static"]
 
 [badges]
 maintenance = { status = "actively-developed" }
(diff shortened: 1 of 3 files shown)


filmil:
Thank you for this contribution and the other PRs you sent. Please allow me some time to get around to reviewing them. Thanks for your patience!

SabrinaJewson:
Ah, I just realized that while my distribution has the system ICU compiled with `-fPIC` (so it works perfectly fine with Rust), some distributions don’t (like Ubuntu) — that means static linking won’t work for them, and we’ll have to add another feature flag to enable building from source. So I’ll convert this to a draft for now.

filmil:
OK, let me know what you want to do with the PR once it's ready.

filmil:
Ping?

SabrinaJewson:
Right, sorry, I forgot to update this. So I tried adding automatic building of ICU to the build script via another feature flag, but the thing is that turns out to be _very_ slow and also annoying because it doesn’t give any output during the build process. I think the best course of action here is to merge this for the use case where people _do_ have a prebuilt PIC ICU available, and then figure out what to do about building from source separately.

mpalmer:
This PR landed at a *really* great time, as I'm in need of static-linked ICU in a Rust project.  Thanks for putting it together, @SabrinaJewson.  Do you have any indication, @filmil, of when this PR is likely to make it into a release?

filmil:
I'm afk for the next 3-4 days but and will try to release thereabouts.

Issue #270, opened 2023-05-01
Upgrade to ICU 73.1


filmil:
This is your regular once-in-6-months update of the major ICU version.

ICU 73.1 was released in mid-April 2023: https://icu.unicode.org/download/73

rust_icu sometimes has an issue with change detector tests which get tripped up by evolving CLDR information in some languages.

filmil:
This is done.

PR #271, opened 2023-05-01
Adds ICU v73.1 to the built images

filmil:
At the same time, v70.1 is no longer supported.

I'll try to have the version numbering schema reflect the latest version of ICU that is supported. So, at this time, we will upgrade the buildenv to
`buildenv-1.73.0`.

Issue: #270


filmil:
@Manishearth though there are plenty of files in the PR, the changes should be straightforward.

Manishearth (approved):

PR #277, opened 2023-08-01
Update buildenv to rust 1.71.0

filmil:
This handles the issue with some dependencies dropping support for older rust versions.

See https://github.com/google/rust_icu/pull/274 for details.

Issue: #276

PR #278, opened 2023-08-01
Adds a once-per-week scheduled run for the build

filmil:
This might allow us to catch bitrot issues early.

Bitrot is fixed in https://github.com/google/rust_icu/pull/277, so this probably needs to wait until that PR is merged.


ComputerDruid (approved):

filmil:
Thanks! Now merged.

PR #284, opened 2023-10-06
chore: remove statements causing warnings

filmil:
Remove statements which, for some reason, caused older compiler versions to error out.

This doesn't seem to be the case now, and there should be no reason to keep them.


filmil:
@Manishearth ping?

Manishearth (approved):

PR #287, opened 2023-10-23
Move linking flags to build.rs to help with non-cargo build systems

gferon:
While using `rust_icu` with Bazel, I realized that the `#[cfg_attr]` attributes in the code make everything difficult.

This brings the following improvement:
- Makes it easier to add combination of features to change the flags, since we can use proper branching
- Moving the instructions to the `build.rs` enables building `rust_icu` with Bazel (which can skip using the `build.rs` file)
- Adds the ability to specify the `RUST_ICU_LINK_SEARCH_DIR` environment variable when you need to use a non-installed build of `icu4c`
- Document the `static` feature and add a CI job to test this for both Ubuntu and macOS

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,3 +46,14 @@ jobs:
       - uses: actions/checkout@v2
       - name: 'Test static-bindgen'
         run: 'make static-bindgen'
+  test-static-linking:
+    runs-on: ${{ matrix.runs-on }}
+    strategy:
+      matrix:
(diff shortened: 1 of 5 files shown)


filmil:
Please rebase, and let me know when to continue reviewing.

gferon:
This is ready for a review now!

filmil:
Thank you for the contribution, for improving the linking recipes, and for the new workflow configuration!

All are most welcome.

PR #290, opened 2023-10-25
Set the resolver version to 2, clarify the error message when the ICU…

clydegerber:
… version isn't found and add implementations for ICU functions:
 - uloc_getName(),
 - uloc_getISO3Language(),
 - uloc_getISO3Country(),
 - uloc_getDisplayLanguage(),
 - uloc_getDisplayScript(),
 - uloc_getDisplayCountry(),
 - uloc_getDisplayVariant(),
 - uloc_getDisplayKeyword(),
 - uloc_getDisplayKeywordValue(), and
 - uloc_getDisplayName()).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,4 @@ members = [
   "rust_icu_utext",
   "rust_icu_utrans",
 ]
+resolver = "2"
--- a/rust_icu_sys/build.rs
+++ b/rust_icu_sys/build.rs
@@ -225,7 +225,7 @@ mod inner {
                         Ok(str)
                     }
                 })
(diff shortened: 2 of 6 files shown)


filmil:
Your new tests are failing likely because the old ICU data does not have all the locale information present in more modern ICU versions, causing different results.

You may want to add a config flag on the value of `icu_version_67_plus` on the affected tests, causing them to be skipped when compiled with ICU v63.

https://github.com/google/rust_icu/blob/c6dac109bfd10645381e8ceeb23b2de86cc9e7e8/rust_icu_ecma402/src/listformat.rs#L30

(In general I recommend depending less on the *exact* text in the ICU data responses as ICU data tends to change frequently. But I admit this is something I did early on too.)

PR #292, opened 2023-10-27
Provide a new API that returns the vector of available ICU locales us…

clydegerber:
…ing uloc_countAvailable() and uloc_getAvailable().

### DIFF
--- a/rust_icu_uloc/src/lib.rs
+++ b/rust_icu_uloc/src/lib.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use {
-    once_cell::sync::OnceCell,
     rust_icu_common as common,
     rust_icu_common::buffered_string_method_with_retry,
     rust_icu_sys as sys,
@@ -28,6 +27,7 @@ use {
         convert::{From, TryFrom, TryInto},
         ffi, fmt,
         os::raw,
+        sync::OnceLock,
     },
 };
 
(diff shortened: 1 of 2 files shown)


rust_icu_uloc/src/lib.rs
```diff
@@ -1244,4 +1262,9 @@ mod tests {
             "azerbaïdjanais (cyrillique, Azerbaïdjan, calendrier=calendrier hébraïque, t=it, usage privé=whatever)"
         );
     }
+    #[test]
+    fn test_get_available_locales() {
```
filmil, line 1266: Is there any fairly stable test for the contents of the returned vectors that you could add?

rust_icu_uloc/src/lib.rs
```diff
@@ -361,6 +362,23 @@ impl ULoc {
         )
     }
 
+    /// Returns the collection of available locales via 'uloc_countAvailable'/'uloc_getAvailable'
```
filmil, line 365: Add a formulaic /// Implements `...` here, for coverage.

Also if you are able can you run `make cov` to update the coverage data?
`make cov` is not hermetic and requires `exuberant-ctags`. If you don't have that package or can't install it, feel free not do it.

rust_icu_uloc/src/lib.rs
clydegerber: I should be able to test for the existence of a few well-known locales. Will investigate.

rust_icu_uloc/src/lib.rs
clydegerber: Will add the 'Implements' comment and see if I can run the 'make cov'

clydegerber: I installed exuberant-ctags but the output in my dev environment (Mac OS) doesn't look correct, so not including that.

PR #293, opened 2023-10-30
Replace the whitespace character in the sed regular expression that o…

clydegerber:
…perates on ctags output with a space and replace 'echo' with 'printf' to write out the coverage report. Sort output files with a stable, case-insensitive sort.


clydegerber:
These changes were required to get 'make cov' to produce reasonable output on a MacOS. Open to any suggestions - I think that replacing the '\s' with a ' ' in the regular expression should be safe, as it is operating on ctags output. The MacOS sed apparently doesn't recognize '\s' in a sed reg ex and was treating it as a literal. 'echo' was introducing some additional spaces which I've eliminated by replacing it with 'printf'. I don't understand why the files in which some functions were re-ordered were as they were before unless there was some previous manual editing(?) - the '_' character doesn't look like it was sorted consistently previously.

filmil:
Thanks for the contribution.

PR #295, opened 2023-10-31
feat: Implement uloc_openAvailableByType()

clydegerber:
- Change the signature of ULoc::get_available() to more closely match the ICU4C uloc_getAvailable() function signature.
- Add the function ULoc::get_available_locales() to return an (uncached) collection of available locales.
- Add the function ULoc::get_available_locales_by_type() to return a collection of available locales for the specified type.
- Correct some configuration tags that used the plural 'features' incorrectly and were not functioning correctly as a result.


rust_icu_uloc/src/lib.rs
```diff
@@ -385,6 +385,36 @@ impl ULoc {
...
+        let mut status = common::Error::OK_CODE;
+        unsafe {
+            let iter = Enumeration::from_raw_parts(None, versioned_function!(uloc_openAvailableByType)(locale_type, &mut status));
+            common::Error::ok_or_warning(status)?;
+            Ok(iter)
+        }
+    }
+
+    /// Returns a vector of locales of the requested type.
+    pub fn get_available_by_type(locale_type: ULocAvailableType) -> &'static Vec<ULoc> {
```
filmil, line 399: It seems that more of this functionality will need to be pushed down to ICU.

clydegerber:
Looks file test failures against the 6.3 ICU version were due to the uloc_openAvailableByType() API not being available until 6.5. I've attempted to add a feature for the 6.5 version and make the new API's dependent on that feature. I would appreciate a thorough review of what I've done, as I simply followed the examples for the other version features. In the course of this I noticed an issue in my previous commits that I'm correcting as well.

rust_icu_uloc/src/lib.rs
clydegerber: Are you okay with adding the static vectors to this library? The underlying assumption is that the ICU library is loaded once so that it is safe to cache these static results. If dynamically opening and closing the ICU library underneath this library is required then they may not be a good idea.

filmil: Indeed, assuming static data is not going to work. ICU data can be loaded and unloaded.

In other places we've used this was to make it into iterator using Enumerate, but it seems that this particular API doesn't follow that pattern.

It might be OK to recompute the vector on every call, and have the user cache the results instead if they need to.

clydegerber: Removed the caching and recomputing on every call - have a look.

filmil:
Will do, but please allow a day or so.

F

filmil:
Sorry - I had these comments pending for a while but didn't send them out.

PR #296, opened 2023-11-05
fix: Improve the accuracy of the coverage report.

clydegerber:
- Ignore trailing '()' characters in function names when reporting coverage.
- Only track functions that match a function in the header file of interest as implemented.
- Mark the uenum::uloc_open_keywords as deprecated, copy the implementation into the uloc crate and copy the test that was in the     uloc crate into the uenum crate so that it is still protected by a test even though deprecated.

### DIFF
--- a/coverage/report.md
+++ b/coverage/report.md
@@ -5,18 +5,18 @@
 | `ubrk.h` | 19 / 23 | 
 | `ucal.h` | 15 / 47 | 
 | `ucol.h` | 8 / 51 | 
-| `udat.h` | 10 / 38 | 
-| `udata.h` | 4 / 8 | 
-| `uenum.h` | 7 / 8 | 
+| `udat.h` | 5 / 38 | 
+| `udata.h` | 2 / 8 | 
+| `uenum.h` | 3 / 8 | 
 | `uformattable.h` | 6 / 13 | 
 | `ulistformatter.h` | 2 / 8 | 
 | `uloc.h` | 28 / 42 | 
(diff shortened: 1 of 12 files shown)


coverage/report.md
```diff
@@ -205,6 +205,13 @@
 | `udat_toPatternRelativeDate` | |
 | `udat_toPatternRelativeTime` | |
 | `udat_unregisterOpener` | |
+Warning: Implemented + Unimplemented != Total
```
filmil, line 208: ?

filmil:
Is it possible to use the canonical git commit messages, with a short one-line summary, and a slightly longer synopsis?

E.g. like https://www.conventionalcommits.org/en/v1.0.0/#summary - though I'm not a stickler to the specific format.

clydegerber:
> Is it possible to use the canonical git commit messages, with a short one-line summary, and a slightly longer synopsis?
> 
> E.g. like https://www.conventionalcommits.org/en/v1.0.0/#summary - though I'm not a stickler to the specific format.

Can do!

coverage/report.md
clydegerber: Is the question about the clarity of the message or the existence of the message? If it is about clarity, would something like "Warning: Count of total C API functions is not equal to the sum of the counts of implemented and non-implemented functions" be clearer?

clydegerber: An alternative would be to add a screen to the '*_implemented.txt' file creation to ignore functions that aren't in the C API for the given header file.

Issue #305, opened 2024-11-08
Build error on Fedora 41


vmiklos:
Hi,

I ran into a weird problem while using the `rust_icu_ucol` crate. Fedora 40 with rustc 1.82.0 compiles this just fine:

- Cargo.toml:

```
[package]
name = "test"
version = "0.1.0"
edition = "2021"
[dependencies]
rust_icu_ucol = "5.0.0"
```

- src/main.rs:

```
fn main() {
}
```

And then on Fedora 40 `cargo run` passes fine. (After `dnf install libicu-devel` and `dnf install clang-devel`.)

The same fails on Fedora 41 with:

```
# cargo run
   Compiling rust_icu_ustring v5.0.0
   Compiling rust_icu_uenum v5.0.0
error[E0425]: cannot find function `u_strFromUTF8_74` in crate `$crate`
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rust_icu_ustring-5.0.0/src/lib.rs:161:13
    |
161 |             versioned_function!(u_strFromUTF8)(
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `$crate`
    |
    = note: this error originates in the macro `versioned_function` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find function `u_strFromUTF8_74` in crate `$crate`
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rust_icu_ustring-5.0.0/src/lib.rs:180:13
    |
180 |             versioned_function!(u_strFromUTF8)(
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `$crate`
    |
    = note: this error originates in the macro `versioned_function` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find function `u_strToUTF8_74` in crate `$crate`
   --> /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rust_icu_ustring-5.0.0/src/lib.rs:210:13
    |
210 |             versioned_function!(u_strToUTF8)(
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `$crate`
    |
    = note: this error originates in the macro `versioned_function` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0425]: cannot find function `u_strToUTF8_74` in crate `$crate`
[...]

PR #306, opened 2024-11-09
rust_icu_sys: fix build with clang-19

vmiklos:
Building in a Fedora 41 container againts clang-19 results in a lib.rs (output from bindgen) that is missing key functions like u_strFromUTF8.

Digging deeper, this is a problem with bindgen, see <https://github.com/rust-lang/rust-bindgen/issues/2945>.

Update bindgen to a newer version which is just new enough to fix the problem.

Fixes <https://github.com/google/rust_icu/issues/305>.

### DIFF
--- a/rust_icu_sys/Cargo.toml
+++ b/rust_icu_sys/Cargo.toml
@@ -22,7 +22,7 @@ paste = "1.0"
 
 [build-dependencies]
 anyhow = "1.0.72"
-bindgen = { version = "0.59.2", optional = true }
+bindgen = { version = "0.69.5", optional = true }
 lazy_static = "1.4"
 
 [lib]


vmiklos:
Thanks for the review. :-)

vmiklos:
@filmil do you think you could cut a new release in the near future? If I see correctly, the last release was in February, and the last release doesn't build on stable Fedora till this PR hits a release. Thanks. :-)

filmil:
I recommend against using release tags for pinning the source. Please use
commit IDs as those are immutable.

vmiklos:
I didn't mean git tags, I meant a release to crates.io. Or do you discourage the usage of those releases, too? :-) Thanks.

PR #311, opened 2024-11-13
doc: update the version support matrix

filmil:
- Retired rust_icu 3.0.
- Retired support for ICU 70 to 73.
- Added a notice that rust_icu compiles well under 76.0, given the new compile workflow.
- We skipped ICU 75, but I'm fairly confident it would have worked.


filmil:
Thank you for your reviews and patience. This is the last PR in this batch, for a while hopefully.

Manishearth (approved):

PR #339, opened 2026-04-15
fix: run bindgen natively on macOS instead of using Linux-generated bindings

### LINKED ISSUE #301: Re-enable test-static-linking macos-latest
Disabled since it seems to have issues with statically linking ICU 74, and I don't have a mac to test this properly.

clydegerber:
## Summary

- Switches `macos-test` from `icu_version_in_env` (pre-built Linux bindings) to native bindgen via `use-bindgen,icu_config,renaming`, eliminating a dependency on Linux-generated artifacts that embedded Linux-specific type definitions
- Adds a second `macos-test` pass with the `static` feature to exercise static linking on macOS; fixes a `build.rs` conflict that previously made `use-bindgen+icu_config+static` fail with "overriding linking modifiers" — `icu_config_autodetect()` was emitting `cargo:rustc-flags` with dynamic `-l` flags at the same time as the `static` feature's `cargo:rustc-link-lib=static` directives
- Re-enables `macos-latest` in the `test-static-linking` CI job (Closes #301)
- Fixes `rust_icu_utext::test::partial_eq` to use a shallow clone instead of a deep clone: `utext_equals` compares the `context` pointer, which differs after a deep clone (new storage allocated); this test was never run on macOS before because the CI job was disabled
- Adds `rust_icu_utext::test::try_clone_combinations` covering all four `(deep, readonly)` combinations with commentary explaining the ICU source behavior and the uncertainty around whether the original deep-clone assertion was silently failing on Linux

## Test plan

- [x] `make macos-test` passes locally (dynamic and static passes, 2 tests in `rust_icu_utext`)
- [ ] `test-static-linking` CI job passes on both `ubuntu-latest` and `macos-latest`
- [ ] Existing Docker-based CI jobs (`test-default-features`, `test-with-features`) unaffected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,12 +89,26 @@ docker-test-current:
 			${DOCKER_REPO}/rust_icu_testenv-current:${USED_BUILDENV_VERSION}
 .PHONY: docker-test-current
 
-# Test with the homebrew version of icu4c on macOS with static linking (the default way of linking for distribution on Apple platforms)
+# Test with the homebrew version of icu4c on macOS, running bindgen natively
+# using the ICU headers provided by Homebrew and the clang toolchain from Xcode.
+# Two passes: dynamic linking (default) and static linking.
+#
+# brew --prefix icu4c resolves the keg-only formula prefix.  On some
+# environments (e.g. GitHub Actions) the unversioned alias is not recognised
+# and brew --prefix returns a non-zero exit; in that case we fall back to
+# querying brew list for the actual installed formula name (e.g. icu4c@78).
 macos-test:
 	brew install icu4c
-	RUST_ICU_LINK_SEARCH_DIR="$(shell brew --prefix)/opt/icu4c/lib" \
-	RUST_ICU_MAJOR_VERSION_NUMBER=${RUST_ICU_MAJOR_VERSION_NUMBER} \
-	cargo test --no-default-features --features=icu_version_64_plus,icu_version_67_plus,icu_version_68_plus,icu_version_in_env,renaming,static
+	ICU_PREFIX=$$(brew --prefix icu4c 2>/dev/null \
+	    || brew --prefix $$(brew list --formula | grep -m1 '^icu4c')) \
+	    && PATH="$$ICU_PREFIX/bin:$$PATH" \
+	       PKG_CONFIG_PATH="$$ICU_PREFIX/lib/pkgconfig:$$PKG_CONFIG_PATH" \
+	       RUSTFLAGS="-L $$ICU_PREFIX/lib" \
+	       cargo test --no-default-features --features=use-bindgen,icu_config,renaming,icu_version_64_plus,icu_version_67_plus,icu_version_68_plus \
+	    && PATH="$$ICU_PREFIX/bin:$$PATH" \
+	       PKG_CONFIG_PATH="$$ICU_PREFIX/lib/pkgconfig:$$PKG_CONFIG_PATH" \
+	       RUSTFLAGS="-L $$ICU_PREFIX/lib" \
+	       cargo test --no-default-features --features=use-bindgen,icu_config,renaming,icu_version_64_plus,icu_version_67_plus,icu_version_68_plus,static
 .PHONY: macos-test
 
 # Refreshes the static bindgen output (contents of ./rust_icu_sys/bindgen) based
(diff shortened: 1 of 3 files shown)


filmil:
Thank you for your contributions so far!

Please be so kind to split this PR into separate PRs, each handling a single change topic, instead of having the changes all sit together.

clydegerber:
I'm open to changing the PR, but I viewed these changes as all related to fixing issue 301. Can you identify which changes should not be part of that fix? The changes in rust_icu_utext/src/lib.rs are to fix a test that is failing in the static link case - it appears that the test was incorrect to begin with and it's a bit mysterious as to how it passes in other platforms. The additional tests are to provide some insight into the behavior of the function.

filmil:
OK, but it's going to take a bit then.

clydegerber:
No worries - thanks!

clydegerber:
I got some insight into the root cause of the problem documented in issue 301, so I'm marking this as a draft until that fix goes in.

rust_icu_utext/src/lib.rs
```diff
@@ -139,10 +139,69 @@ mod test {
...
         assert_ne!(bar, baz);
 
+        // Shallow clone shares the same underlying context pointer, so
+        // utext_equals returns true and the two objects compare as equal.
+        // A deep clone allocates new string storage, changing the context
+        // pointer, which causes utext_equals to return false.
         assert_eq!(
             foo,
-            foo.try_clone(true, true).expect("clone is a success"),
+            foo.try_clone(false, false).expect("clone is a success"),
```
filmil: This seems to be part of  https://github.com/google/rust_icu/pull/342 I suppose?

rust_icu_utext/src/lib.rs
```diff
@@ -139,10 +139,69 @@ mod test {
...
+    // returns true.  A deep clone (deep=true) allocates new storage, changing the context
+    // pointer, so utext_equals returns false.
+    #[test]
+    fn try_clone_combinations() {
+        let src = Text::try_from("hello").expect("source text");
+
+        // (deep=false, readonly=false): shallow, preserves source writability.
+        let c = src.try_clone(false, false).expect("shallow read-write clone");
+        assert_eq!(src, c, "shallow read-write clone equals source");
+        assert_eq!(0i8, unsafe { versioned_function!(utext_isWritable)(src.rep) },
```
filmil: Can you provide a, is writable check that sequesters `unsafe`, and doesnt' require `versioned_function!` use in tests?

Here, and below.

Makefile
```diff
@@ -89,12 +89,22 @@ docker-test-current:
 			${DOCKER_REPO}/rust_icu_testenv-current:${USED_BUILDENV_VERSION}
 .PHONY: docker-test-current
 
-# Test with the homebrew version of icu4c on macOS with static linking (the default way of linking for distribution on Apple platforms)
+# Test with the homebrew version of icu4c on macOS, running bindgen natively
```
filmil, line 92: I'm going to trust you about macos-test as I don't have a mac to verify.

filmil (changes requested):

rust_icu_utext/src/lib.rs
clydegerber: Yes, I've addressed your comment on that PR - I'll republish this PR after that change is in.

rust_icu_utext/src/lib.rs
clydegerber: Will do.

filmil:
You might want to rebase on top of main before continuing with this.

clydegerber:
> You might want to rebase on top of main before continuing with this.

Yes, I'm working on that - I'll move it out of draft status when I think I'm done.

clydegerber:
> > You might want to rebase on top of main before continuing with this.
> 
> Yes, I'm working on that - I'll move it out of draft status when I think I'm done.

I rebased, and left it as two commits in case you have a preference for splitting this into two PRs - one for the changes to run bindgen on MacOS and the other to streamline the partial_eq tests in rust_icu_utext. I will keep this as a draft until PR 342 goes in as we need that fix in order for the tests to be stable.

Makefile
clydegerber: Well, trust but verify (at least with CI). All but one of the failures in the last CI job were due to the invalid partial_eq() test in rust_icu_utext, which PR 342 fixes, but the static linking test did fail. I was building via cargo and didn't spot the issue before.

PR #342, opened 2026-04-17
fix: eliminate use-after-free in rust_icu_utext TryFrom impls

clydegerber:
## Summary

- `utext_openUTF8` stores the caller's pointer in `ut->context` without copying. Both `TryFrom<String>` and `TryFrom<&str>` were passing a raw pointer into the source buffer and then allowing the backing allocation to be dropped, leaving `ut->context` dangling.
- Fix by opening a temporary UText over the source buffer and immediately deep-cloning it (`utext_clone` with `deep=true`), so ICU allocates and owns a copy of the string data (`UTEXT_PROVIDER_OWNS_TEXT`). The returned `Text` is fully self-contained and the struct remains a thin wrapper around `UText` with no additional fields.
- A NUL terminator is appended to the local buffer before the ICU calls because `utf8TextClone` copies `len+1` bytes via `uprv_memcpy`, reading one byte past the declared string length.
- Fix the `try_clone(true, true)` assertion in `partial_eq`: a deep clone changes the context pointer so `utext_equals` returns false. The assertion was previously passing on Linux only due to glibc tcache LIFO reuse coincidentally returning the freed source address for the subsequent `uprv_malloc` call. Changed to `try_clone(false, false)` (shallow clone), which preserves the context pointer and correctly compares as equal.

## Test plan

- [ ] Existing `partial_eq` test passes with corrected assertion
- [ ] CI passes on Linux (docker-test) and macOS (macos-test)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust_icu_utext/src/lib.rs
+++ b/rust_icu_utext/src/lib.rs
@@ -43,36 +43,51 @@ impl TryFrom<String> for Text {
     ///
     /// The conversion may fail if the string is not well formed, and may result in an error.
     ///
-    /// Implements `utext_open` from ICU4C.
+    /// Implements `utext_openUTF8` from ICU4C.
     fn try_from(s: String) -> Result<Self, Self::Error> {
-        let len: i64 = s.len() as i64;
-        let bytes = s.as_ptr() as *const raw::c_char;
-        // bytes and len must be compatible.  Ensured by the two lines just above.
-        unsafe { Self::from_raw_bytes(bytes, len) }
+        let len = s.len() as i64;
+        let mut bytes = s.into_bytes();
+        bytes.push(0); // NUL terminator; utext_clone deep reads len+1 bytes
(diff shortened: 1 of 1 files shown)


rust_icu_utext/src/lib.rs
```diff
@@ -139,9 +154,13 @@ mod test {
...
 
+        // Shallow clone shares the same underlying context pointer, so
+        // utext_equals returns true and the two objects compare as equal.
+        // A deep clone allocates new string storage, changing the context
+        // pointer, which causes utext_equals to return false.
         assert_eq!(
             foo,
-            foo.try_clone(true, true).expect("clone is a success"),
+            foo.try_clone(false, false).expect("clone is a success"),
             "a clone should be the same as its source"
```
filmil: I suppose that this is now not correct.

filmil (changes requested):

PR #354, opened 2026-04-29
🔒 [security fix] Fix unwrap() panics when converting C strings

filmil:
🎯 **What:** The vulnerability fixed is the use of `unwrap()` when converting C strings to Rust strings in `rust_icu_uloc`.

⚠️ **Risk:** If the underlying ICU library returns a string that is not valid UTF-8, the current code would panic and crash the application.

🛡️ **Solution:**
- In `ULoc::get_available`, `unwrap()` is replaced with the `?` operator to propagate any UTF-8 conversion errors.
- In `ULoc::iso3_language` and `ULoc::iso3_country`, `to_str().unwrap()` is replaced with `to_string_lossy()` to avoid panics while maintaining the existing API signature.

These changes ensure the library is more robust against unexpected input from the underlying C library.

---
*PR created automatically by Jules for task [17310211343422451087](https://jules.google.com/task/17310211343422451087) started by @filmil*

### DIFF
--- a/rust_icu_uloc/src/lib.rs
+++ b/rust_icu_uloc/src/lib.rs
@@ -226,12 +226,12 @@ impl ULoc {
 
     /// Implements 'uloc_getISO3Language' from ICU4C.
     pub fn iso3_language(&self) -> Option<String> {
-        let lang = unsafe {
-            ffi::CStr::from_ptr(
-                versioned_function!(uloc_getISO3Language)(self.as_c_str().as_ptr())
-            ).to_str()
+        let value = unsafe {
+            ffi::CStr::from_ptr(versioned_function!(uloc_getISO3Language)(
+                self.as_c_str().as_ptr(),
+            ))
+            .to_string_lossy()
         };
-        let value = lang.unwrap();
         if value.is_empty() {
             None
         } else {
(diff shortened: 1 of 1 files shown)


clydegerber:
test cases?

rust_icu_uloc/src/lib.rs
```diff
@@ -226,12 +226,12 @@ impl ULoc {
...
     pub fn iso3_language(&self) -> Option<String> {
-        let lang = unsafe {
-            ffi::CStr::from_ptr(
-                versioned_function!(uloc_getISO3Language)(self.as_c_str().as_ptr())
-            ).to_str()
+        let value = unsafe {
+            ffi::CStr::from_ptr(versioned_function!(uloc_getISO3Language)(
+                self.as_c_str().as_ptr(),
+            ))
+            .to_string_lossy()
```
clydegerber: Are we better off breaking the API and returning a Result from these two functions?

rust_icu_uloc/src/lib.rs
```diff
@@ -226,12 +226,12 @@ impl ULoc {
 
     /// Implements 'uloc_getISO3Language' from ICU4C.
     pub fn iso3_language(&self) -> Option<String> {
-        let lang = unsafe {
-            ffi::CStr::from_ptr(
-                versioned_function!(uloc_getISO3Language)(self.as_c_str().as_ptr())
-            ).to_str()
+        let value = unsafe {
```
filmil: Add tests, for this and below.

rust_icu_uloc/src/lib.rs
filmil: Not at this time.

PR #356, opened 2026-04-29
fix: address compiler warnings across crates

clydegerber:
## Summary

- `rust_icu_udata`: adds `#[allow(dead_code)]` to `Rep::Buffer(Vec<u8>)` — this variant intentionally keeps ICU's raw pointer alive and is never read directly
- `rust_icu_ucnv`: fixes copy-paste bug in pivot buffer reset functions where `pivot_to_target` and `pivot_from_target` were each assigned to themselves instead of their respective `start` pointers
- `rust_icu_ucsdet`: makes the anonymous lifetime explicit in `detect()` and `detect_all()` return types (`CharsetMatch` → `CharsetMatch<'_>`)
- `rust_icu_common`: adds `#[allow(dead_code)]` to the `Type` struct in the hygiene test, which exists solely to exercise `simple_drop_impl!` under `#![no_implicit_prelude]` and is never constructed

## Test plan

- [ ] CI passes with no compiler warnings

### DIFF
--- a/rust_icu_common/tests/hygiene.rs
+++ b/rust_icu_common/tests/hygiene.rs
@@ -1,5 +1,6 @@
 #![no_implicit_prelude]
 
+#[allow(dead_code)]
 struct Type {
     rep: ::std::ptr::NonNull<::rust_icu_sys::UMessageFormat>,
 }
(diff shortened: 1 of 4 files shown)


filmil:
These are fairly small changes so I'll review them together.

But, please, in the future, split separate fixes across PRs.

PR #363, opened 2026-05-18
fix: add build.rs to crates with version-gated code lacking auto-detection

clydegerber:
## Summary

Adds `build.rs` to `rust_icu_ecma402`, `rust_icu_ulistformatter`, and `rust_icu_unumberformatter`. These crates have `#[cfg(feature = "icu_version_XX_plus")]`-gated code but lacked their own `build.rs`, so they relied on downstream feature propagation rather than ICU auto-detection from `pkg-config`.

The new `build.rs` files follow the same pattern as `rust_icu_uloc/build.rs`, calling `rust_icu_release::run()` to emit `cargo:rustc-cfg` flags based on the installed ICU version.

Also:
- Adds a "ICU version feature detection" section to `CONTRIBUTING.md` documenting how version-gated code is wired up.
- Simplifies the `macos-test` Makefile target, since the affected crates now self-configure.

## Test plan

- [x] `cargo test -p rust_icu_ulistformatter` runs the previously-skipped version-gated test
- [ ] CI passes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,47 @@ term health of the repository.
 * Please keep PRs limited to a single topic of change. This makes the PR easier to
   review, and easier to roll back, if that becomes necessary.
 
+## ICU version feature detection
+
+Several crates in this workspace contain code that is conditionally compiled
+based on the ICU library version, using attributes such as
+`#[cfg(feature = "icu_version_64_plus")]`.
+
+**There are two mechanisms by which these features are activated, and both
+must be kept in sync:**
+
+### 1. Automatic detection via `build.rs` (used when `icu_config` is active)
+
+Crates that have a `build.rs` call `rust_icu_release::run()`, which queries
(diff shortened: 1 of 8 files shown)


filmil:
I'm not sure what to think about the proliferation of `build.rs`. But I do admit that it resolves a headache.

Flip side, most `XX` values are now *smaller* than the minimum ICU version we support, so maybe we can instead just remove those feature flags.

Thoughts?

clydegerber:
Removing the obsolete feature flags is a good call. I'll work it up.

clydegerber:
I created a new PR (364) for this approach. I can close whichever one we don't want to go with.

clydegerber:
Closing in favor of #364, which addresses the same root issue (version-gated code that no longer needs gating) by removing the dead conditional-compilation mechanism rather than adding build scripts to keep it alive. The minimum supported ICU version is now 74, so the gates can never be false — #364 removes the dead code rather than wiring it up.

PR #366, opened 2026-06-03
docs: add rust_icu hello world example

arnavnagzirkar:
Adds a concise Hello World section to the README showing how to use rust_icu_umsg MessageFormat with rust_icu_common, rust_icu_uloc, and rust_icu_ustring dependencies.\n\nThis is a docs-only change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,15 @@ jobs:
       - uses: actions/checkout@v6
       - name: 'Test ICU version ${{ matrix.icu_version }}'
         run: 'make DOCKER_TEST_ENV=rust_icu_testenv-${{ matrix.icu_version}} docker-test'
+  test-examples:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        icu_version: [74, 76, 77]
+    steps:
+      - uses: actions/checkout@v6
+      - name: 'Build and run examples on ICU version ${{ matrix.icu_version }}'
+        run: 'make DOCKER_TEST_ENV=rust_icu_testenv-${{ matrix.icu_version }} RUST_ICU_MAJOR_VERSION_NUMBER=${{ matrix.icu_version }} docker-test-example'
   test-with-features:
     runs-on: ubuntu-latest
     strategy:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,11 @@ members = [
   "rust_icu_utext",
   "rust_icu_utrans",
 ]
+# Examples are standalone crates that build against the workspace crates via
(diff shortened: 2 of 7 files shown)


filmil:
How about an actual "examples subdir with the configurations you want to demo?

arnavnagzirkar:
Good call — done. I moved the demo into a standalone, runnable `examples/hello_world` crate:

- `examples/hello_world/Cargo.toml` — the exact dependency configuration to demo (`rust_icu_common`, `rust_icu_uloc`, `rust_icu_umsg`, `rust_icu_ustring`), with a comment showing the crates.io version alternative for downstream users.
- `examples/hello_world/src/main.rs` — `cargo run` prints `Hello, World!`.
- `examples/hello_world/README.md` — how to run it.

The crate is added to `exclude` in the root `Cargo.toml` so the CI feature matrix (`cargo test --no-default-features --features ...`) is unaffected, and the top-level README now links to the example instead of inlining the code.

filmil:
Is this something that could be tied into presubmits? It would be much more useful if it's actually checked that it works.

arnavnagzirkar:
Good idea. I have wired the example into the presubmits so it is actually verified on every push and pull request.

What I changed:

* `examples/hello_world/src/main.rs` now puts the formatting logic in a `greet()` function with a `#[test]` that asserts the output equals `Hello, World!`, so the check verifies correctness and not just that the code compiles.
* A new `docker-test-example` target in the `Makefile` builds and runs the example inside the same dockerized test environment the rest of the suite uses. It runs `cargo test` (the assertion) and then `cargo run`.
* A new `test-examples` job in `.github/workflows/test.yml` runs that target across ICU 74, 76 and 77, matching the existing default feature matrix.

Because the example depends on the in tree crates with their default features, I kept it on the default feature configuration rather than the `--no-default-features --features ...` matrix, which targets the library crates. If the example ever stops compiling or stops printing `Hello, World!`, CI now fails.

You can reproduce the presubmit locally with:

```sh
make DOCKER_TEST_ENV=rust_icu_testenv-77 RUST_ICU_MAJOR_VERSION_NUMBER=77 docker-test-example
```

filmil:
Hi @arnavnagzirkar, this PR looks amazing!
Since the project has recently migrated to Bazel-based CI pipelines for tests instead of raw docker scripts, the PR needs a small `BUILD.bazel` for the `examples/hello_world` crate to build successfully.

I've pushed a branch that contains this `BUILD.bazel` fix to the repository under `issue-366-bazel`. Could you checkout this branch or cherry-pick commit `d58d3953270e154c9f5428a0a68d4da5f395dd0a` to append the bazel configuration to your worktree? Thanks!

Issue #371, opened 2026-07-28
Segfault when using `message_format!` on an invalid message format


Mistera91:
Example :

```rust
use rust_icu_common as common;
use rust_icu_uloc as uloc;
use rust_icu_umsg as umsg;
use rust_icu_ustring as ustring;

fn main() -> Result<(), common::Error> {
    let loc = uloc::ULoc::try_from("en-US-u-tz-uslax")?;
    let msg = ustring::UChar::try_from(r"String : {1}")?; // notice how there is no `{0}` in the string

    let fmt = umsg::UMessageFormat::try_from(&msg, &loc)?;
    let string = ustring::UChar::try_from("Hello!")?;
    let result = umsg::message_format!(
      fmt,
      { string => String },
    )?; // segfault here

    println!("{result}");
    Ok(())
}
```

Not sure if it’s the fault of the wrapper or ICU’s `umsg`.

PR #372, opened 2026-07-30
feat(umsg): add unsafe try_format method surfacing umsg_format's contract

clydegerber:
## Summary

Addresses the segfault reported in #371 **without breaking the existing API**, per review feedback.

`message_format!` expands to a call into the ICU C *variadic* function `umsg_format`, which derives the number and types of the arguments it reads from the **pattern** passed to `UMessageFormat::try_from`, not from the arguments actually supplied. Nothing reconciles the two, so a mismatch (wrong argument count or wrong argument type) is undefined behavior — segfault or silent memory corruption — reachable from safe-looking code.

The previous revision of this PR made `message_format!` itself `unsafe`. As @filmil noted, that breaks existing callers. This revision instead leaves `message_format!` **unchanged and fully backward compatible**, and adds an opt-in alternative that surfaces the contract.

## Changes (purely additive)

- **New `UMessageFormat::try_format`** — an `unsafe fn` taking a tuple of arguments (`f64`/`i32`/`i64`/`UChar`). It carries the same contract as the macro but forces the caller to acknowledge it with an `unsafe { .. }` block:
  ```rust
  // SAFETY: pattern references one double argument, matching the tuple.
  let result = unsafe { fmt.try_format((43.4_f64,)) }?;
  ```
- **Documentation** — a `# Safety` section on `try_format` and a `# Warning: memory-safety hazards` section on `message_format!`, both spelling out the argument-count and argument-type invariants and linking #371.
- **Tests** — a positive test for `try_format`, plus two `#[ignore]`d demonstrator tests reproducing the count and type mismatches through `message_format!` (kept out of CI because they trigger UB).

`message_format!`, `format_args`, and `checkarg!` are byte-for-byte unchanged relative to `main`; the diff is +207/−2 (the two deletions are a superseded doc note).

## Note on platform-dependence

Observed when running the ignored tests on macOS with ICU 73:
- **type mismatch** → `SIGSEGV`;
- **arg-count mismatch** → does *not* crash here; silently returns garbage (`"String : "`).

So #371's segfault is platform/ICU-version-dependent (same shape seen previously in `rust_icu_utext`). Either outcome is unsound.

## Testing

- `cargo test -p rust_icu_umsg -p rust_icu_intl`: unit + doc tests pass; the two UB tests are ignored; no warnings.

Refs #371

This commit was created by an automated coding assistant, with human
supervision.

### DIFF
--- a/rust_icu_umsg/src/lib.rs
+++ b/rust_icu_umsg/src/lib.rs
@@ -177,6 +177,70 @@ impl UMessageFormat {
             rep: std::rc::Rc::new(Rep { rep }),
         })
     }
+
+    /// Formats `args` into this formatter's message, returning the formatted string.
+    ///
+    /// This is the explicit-`unsafe` counterpart to the [message_format!] macro: it carries exactly
+    /// the same contract, but as an `unsafe fn` it forces the caller to acknowledge that contract
+    /// with an `unsafe { .. }` block. Prefer it over the macro when you want the obligation to be
+    /// visible at the call site.
+    ///
+    /// `args` is a tuple of the values to format, one per positional parameter, in order. Each
+    /// element's type selects how ICU reads it (see the type table on [message_format!]):
+    ///
+    /// | Tuple element type | MessageFormat role |
+    /// | ------------------ | ------------------ |
+    /// | `f64` | `Double` (and `Date`, since [rust_icu_sys::UDate] is an `f64`) |
+    /// | `i32` | `Integer` |
+    /// | `i64` | `Long` |
+    /// | [rust_icu_ustring::UChar] | `String` |
+    ///
+    /// For example, `(43.4_f64, 31337_i32)` binds a double to parameter `{0}` and an integer to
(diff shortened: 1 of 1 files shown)


filmil:
This is not OK, since it breaks existing APIs. We have code that would break as a result of this change.
Please introduce a new formatter if you want to emphasize unsafety.

clydegerber:
Thanks for the review, reverting the breaking change.

`message_format!` is now byte-for-byte identical to `main` (all call sites, including `rust_icu_intl::PluralRules::select`, are back to what they were). Instead I've added an opt-in `UMessageFormat::try_format` — an `unsafe fn` that takes a tuple of arguments and carries the same contract, so callers who want the unsafety surfaced can use:

```rust
let result = unsafe { fmt.try_format((43.4_f64,)) }?;
```

The `# Safety`/`# Warning` docs and the `#[ignore]`d demonstrator tests are kept, since they document the hazard regardless of which entry point is used.

Since this PR does **not** actually stop the #371 segfault I've changed the PR to say "Refs #371" rather than "Closes" for that reason.

This comment was created by an automated coding assistant, with human
supervision.

Issue #376, opened 2026-08-11
Use `bazel` to build and release rust_icu


filmil:
Using bazel to build and release `rust_icu` should remove the remaining manual toil in releases, and ensure timely availability of any
new changes, including changes that must be made on short notice, if any.

This change should do away with the bespoke build machinery we used in the past, which required docker builds. Removing docker environments will remove a significant part of the release process complexity.

Flip side, would-be contributors might be less familiar with `bazel` as a build system. However, from personal experience I believe that a short guide will be able to fill any knowledge gaps fairly quickly, and that not having to deal with the build mechanics will be a win.

PR #378, opened 2026-08-11
build(bazel): Generate BUILD.bazel for all rust_icu crates

filmil:
This PR auto-generates \`rust_library\` and \`rust_test\` integrations for all 25 higher-level crates, resolving all static ICU linkage data.

All tests are verified to pass natively under Bazel sandbox through direct compilation.

Issue: https://github.com/google/rust_icu/issues/376


filmil:
@Manishearth all the BUILD.bazel files should be of exactly the same form in this change. (With the exception of crate-specific naming and such.)

Manishearth (approved):

PR #398, opened 2026-08-12
docs: use reference style links in tables and wrap to 80 cols

filmil:
Reflows README to 80 cols per standard markdown formatting. Offloads raw URLs in tables using numerical disambiguators to vastly reduce horizontal table width footprint.


filmil:
The build error is unrelated.

PR #401, opened 2026-08-12
fix(umsg): null terminate all string arguments to umsg_format

filmil:
umsg_format is variadic, meaning it relies on strings being NUL-terminated. When placing a ustring::UChar object into a message_format! macro for a String argument, the string is not automatically null terminated (unlike how they are typically used in safe rust context). If a string isn't null terminated, umsg_format reads until it encounters a null byte in memory, leading to buffer overflows and undefined behavior (e.g., returning U_BUFFER_OVERFLOW_ERROR when the intended buffer was actually large enough to hold the exact string). 

This properly null terminates all static strings passed to message_format! macro invocations in tests and documentation examples. Fixes the failed CI test on ICU main (77) that was caused by this exact issue.

### DIFF
--- a/rust_icu_umsg/src/lib.rs
+++ b/rust_icu_umsg/src/lib.rs
@@ -62,7 +62,8 @@
 //!     )?;
 //!
 //!     let fmt = umsg::UMessageFormat::try_from(&msg, &loc)?;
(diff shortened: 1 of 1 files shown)


filmil:
The builder error is unrelated.

